PR #22, opened 2015-02-18
(PUP-4026) Specify puppet install paths

joshcooper:
Consolidate information from various tickets (PUP-3809, PUP-3632,
RE-3976 and confluence).

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ Index
   * [Plugin Ruby API][12] - the API for plugins in Ruby
 * General
   * [Settings and Options][13] - specification of settings and options 
+  * [Puppet Installation Layout][15] - specification of Puppet related files on disk
 
 [2]:language/intro.md
 [3]:language/modus-operandi.md
@@ -69,5 +70,6 @@ Index
 [12]:language/plugin-api.md
 [13]:language/settings.md
 [14]:language/puppet-functions.md
+[15]:file_paths.md
 
 [1]:http://www.github.com/puppetlabs/puppet
--- a/file_paths.md
+++ b/file_paths.md
@@ -1,0 +1,155 @@
+Puppet Installation Layout
+==========================
+
+This table specifies the file paths in a Puppet installation and the corresponding settings, denoted as `:setting`. The 3.x column shows the difference of the specification vs the 3.x implementation.
+
+puppet-agent
+------------
+
+    Path                                  Setting                        3.x
+    /etc/puppetlabs                                                      n/a
+
+    /etc/puppetlabs/code                  # :codedir                     contents moved from :confdir
+        environments                      # :environmentpath
+        hiera.yaml                        # :hiera_config
+        hieradata                         # n/a
+        modules                           # user modulepath
+
+    /etc/puppetlabs/mcollective
+        client.cfg
+        facts.yaml
+        server.cfg
+
+    /etc/puppetlabs/puppet                # :confdir                     /etc/puppet
+        auth.conf                         # :rest_authconfig
+        autosign.conf                     # :autosign
+        binder_config.yaml                # :binder_config
+        csr_attributes.yaml               # :csr_attributes
+        custom_trusted_oid_mapping.yaml   # :trusted_oid_mapping_file
+        device.conf                       # :deviceconfig
+        fileserver.conf                   # :fileserverconfig
+        puppet.conf                       # :config
+        routes.yaml                       # :route_file
+        ssl                               # :ssldir                      /etc/puppet/ssl
+
+    /opt/puppetlabs/bin                   # symlink targets of puppet related binaries
+        cfacter@                          -> /opt/puppetlabs/puppet/bin/cfacter
+        facter@                           -> /opt/puppetlabs/puppet/bin/facter
+        hiera@                            -> /opt/puppetlabs/puppet/bin/hiera
+        mco@                              -> /opt/puppetlabs/puppet/bin/mco
+        puppet@                           -> /opt/puppetlabs/puppet/bin/puppet
+
+    /opt/puppetlabs/mcollective/
+        plugins
+
+    /opt/puppetlabs/puppet                # ruby-puppet root
+        bin
+            cfacter
+            facter
+            gem
+            hiera
+            mco
+            mcollectived
+            openssl
+            puppet
+            ruby
+            virt-what
+        cache                             # :vardir                      /var/lib/puppet
+            bucket                        # :bucketdir
+            client_yaml                   # :clientyamldir
+            client_data                   # :client_datadir
+            clientbucket                  # :clientbucketdir
+            devicedir                     # :devices
+            facts.d                       # :pluginfactdest
+            lib                           # :libdir
+            facts                         # used to generate :factpath
+            puppet-module                 # :module_working_dir
+            reports                       # :reportdir
+            server_datadir                # :server_data
+            state                         # :statedir
+            yaml                          # :yamldir
+        include
+            facter
+            openssl
+        lib
+            libaugeas.so
+            libcrypto.so
+            libfacter.so
+            libruby.so
+            libssl.so
+            ruby
+                vendor_ruby               # ruby code
+                    cfacter.rb
+                    facter.rb
+                    hiera.rb
+                    mcollective.rb
+                    puppet.rb
+            virt-what
+                virt-what-cpuid-helper
+        modules                           # system modulepath            /usr/share/puppet/modules
+        share
+            augeas
+            man
+            vim
+        ssl
+
+    /var/log/puppetlabs                   # :logdir                      /var/lib/puppet/log
+        puppet.log                        # not enabled by default
+        mcollective.log
+        mcollectived.log
+
+    /var/run/puppetlabs                   # :rundir                      /var/lib/puppet/run
+        agent.pid                         # :pidfile
+        mcollective.pid
+        mcollectived.pid
+
+These sections describe other Puppet packages that rely on puppet-agent to create the initial directory layout. It does not attempt to specify the full set of file paths for these packages, just cases where the other package has a dependency on puppet-agent.
+
+puppetdb
+--------
+
+    /etc/puppetlabs/puppet
+        puppetdb.conf
+
+
+puppet-server
+-------------
+
+    /etc/puppetlabs/puppetserver
+        logback.xml
+        conf.d
+            puppetserver.conf
+
+    /opt/puppetlabs/bin                   # symlinks of puppet server binaries
+        puppetmaster@                     -> /opt/puppetlabs/server/apps/puppetserver/bin/puppetmaster
+
+    /opt/puppetlabs/server                # serverside apps live underneath
+        apps
+            httpd                         # httpd app dir
+                bin
+                    httpd
+                etc
+                lib
+
+            puppetserver                  # puppetserver app dir
+                bin
+                    puppetmaster
+                etc
+                lib
+        bin                               # symlinks of server binaries
+            httpd@                        -> /opt/puppetlabs/server/apps/httpd/bin/httpd
+            puppetmaster@                 -> /opt/puppetlabs/server/apps/puppetserver/bin/puppetmaster
+
+        data
+            puppetserver                  # :vardir (and $HOME for services that use it)
+                bucket                    # :bucketdir
+                reports                   # :reportdir
+                server_datadir            # :server_data
+                yaml                      # :yamldir
+
+Notes
+-----
+
+### ssldir <a id="note-ssldir"></a>
+The current specification calls for the puppet-agent and puppet-master to continue sharing an `ssldir`. The main reason being the master needs to use the agent's private key when acting as an SSL client. There are issues with this approach, but it's not something
+we are trying to solve now.


file_paths.md
```diff
@@ -0,0 +1,127 @@
...
+puppet-server
+-------------
+
+    /etc/puppetlabs/puppetserver
+        logback.xml
+        conf.d
+            puppetserver.conf
+
+    /opt/puppetlabs/server                # serverside apps live underneath 
+        puppetserver                      # PUP-3089 conflicts with SERVER-357
```
joshcooper: https://tickets.puppetlabs.com/browse/PUP-3809 says:

> /opt/puppetlabs/server <= serverside apps live underneath 
> /opt/puppetlabs/server/puppetserver/

But https://tickets.puppetlabs.com/browse/SERVER-357 says

> /opt/puppetlabs/server/data/puppetserver

file_paths.md
```diff
@@ -0,0 +1,108 @@
...
+puppet-agent
+------------
+
+    Path                                  Setting                        3.x
+    /etc/puppetlabs                       # puppetdir                    n/a
+
+    /etc/puppetlabs/code                  # codedir                      (contents in $confdir)
+        environments                      # environmentpath
+        hiera.yaml                        # hiera_config
+        hieradata                         # n/a
```
joshcooper: @hlindberg @ahpook  is `hieradata` really supposed to be under `code`?

file_paths.md
```diff
@@ -0,0 +1,108 @@
...
+        agent.pid
+        mcollective.pid
+        mcollectived.pid
+
+
+puppetdb
+--------
+
+    /etc/puppetlabs/puppetdb
+        puppetdb.conf
```
joshcooper: @nwolfe @kbarber @ahpook 

https://tickets.puppetlabs.com/browse/PUP-3944 says 

> Make sure puppetdb.conf is under $confdir

But https://tickets.puppetlabs.com/browse/PUP-3809 says:

> /etc/puppetlabs/puppetdb <= puppetdb config dir

README.md
```diff
@@ -54,6 +54,7 @@ Index
   * [Plugin Ruby API][12] - the API for plugins in Ruby
 * General
   * [Settings and Options][13] - specification of settings and options 
+  * [Unified Agent Layout][15] - specification of unified agent layout
```
hlindberg: Small discrepancy on name - we are referring to it as AIO not UAL. Maybe say something else here so reader understands what the document is about - perhaps "Installation Layout on Disk"

file_paths.md
hlindberg: that is were it is now on non windows

file_paths.md
```diff
@@ -0,0 +1,106 @@
+File Paths
+==========
+
+In Puppet 4.0, file paths and related settings will be changing as follows.
```
hlindberg: "This table specifies the file paths in an installation of Puppet and the corresponding setting. The 3x. column shows the difference of this version of the specification vs. the actual 3x implementation."

(It is really not up to the specification to state which version of an implementation that complies with the spec).

file_paths.md
joshcooper: k going with that

file_paths.md
```diff
@@ -0,0 +1,108 @@
...
+puppet-server
+-------------
+
+    /etc/puppetlabs/puppetserver
+        logback.xml
+        conf.d
+            puppetserver.conf
+
+    /opt/puppetlabs/server                # serverside apps live underneath
+        puppetserver                      # PUP-3089 conflicts with SERVER-357
```
joshcooper: @nwolfe https://tickets.puppetlabs.com/browse/PUP-3809 says:

> /opt/puppetlabs/server <= serverside apps live underneath 
> /opt/puppetlabs/server/puppetserver/

But https://tickets.puppetlabs.com/browse/SERVER-357 says

> /opt/puppetlabs/server/data/puppetserver

hlindberg:
I am +1 on this in general (not reviewed that the actual paths are as expected)

file_paths.md
kbarber: Should be `/etc/puppetlabs/puppet/puppetdb.conf`. At least like @joshcooper says, we'll be looking for it in $confdir/puppetdb.conf in our code today. This is the terminus configuration btw, which is why it shares the configuration directory of Puppet.

kbarber: Having said that, I have no idea what we are trying to express here, whether we are just throwing out all of PuppetDB's configuration, or only the bits that Puppet cares about (like the terminus configuration). Either way this is wrong.

file_paths.md
```diff
@@ -0,0 +1,108 @@
...
+            facts.d                       # :pluginfactdest
+            lib                           # :libdir
+            facts                         # used to generate :factpath
+            puppet-module                 # :module_working_dir
+            reports                       # :reportdir
+            server_datadir                # :server_data
+            state                         # :statedir
+            yaml                          # :yamldir
+        include
+        lib
```
MikaelSmith: Cfacter will expect /opt/puppetlabs/puppet/lib/libruby.so

file_paths.md
```diff
@@ -0,0 +1,108 @@
+Puppet Installation Layout
+==========================
+
+This table specifies the file paths in an installation of Puppet and the corresponding Puppet setting, denoted as `:setting`. The 3.x column shows the difference of the specification vs the actual 3.x implementation.
+
+puppet-agent
+------------
+
+    Path                                  Setting                        3.x
+    /etc/puppetlabs                       # :puppetdir                   n/a
```
joshcooper: @ahpook @kylog @cprice404 An earlier version of @nwolfe's PR had the `:puppetdir` setting, but I removed that during my refactoring, because it was only mentioned in the confluence doc but not the other tickets. Also I think the name is confusing, a user could easily think `:puppetdir` refers to `/etc/puppetlabs/puppet`, since it is the directory named `puppet`.

I could add the setting, preferably with a different name, and change `:codedir` and `:confdir` to be relative to it?

Or we could just not define a setting for it?

kylog: I vote that we not define a `puppetdir` setting. Less settings == more better, and people are free to configure `codedir` and `confdir` to suit their needs.

file_paths.md
joshcooper: @kbarber trying to express the bits that Puppet cares about, to make sure the puppet-agent AIO package is creating the correct directories, that other project rely on. Also to raise issues like this, where 3809 and 3944 seem to contradict each other.

file_paths.md
cprice404: Yeah, I was in favor of `puppetdir` before we decided to use 'puppet' as the name of the directory that `confdir` points to, but I agree that it seems like it would be confusing now, given the latest state of all the paths.

file_paths.md
kbarber: @joshcooper gotcha :-).

file_paths.md
joshcooper: k, I will remove the reference to `:puppetdir`

file_paths.md
joshcooper: Using puppet-agent 3a63b02c9d471c7c80b74ddd547539c3812f58f9 (from a day or so ago) on rhel7, I only see static lib, not an `.so`. Will this be changing?

```
# ls -la /opt/puppetlabs/puppet/lib/*ruby*
-rw-r--r--. 1 root root 5203372 Feb 15 19:31 /opt/puppetlabs/puppet/lib/libruby-static.a
```

MikaelSmith: Yes, see https://github.com/puppetlabs/puppet-agent/pull/59.

file_paths.md
```diff
@@ -0,0 +1,123 @@
...
+        puppet.conf                       # :config
+        routes.yaml                       # :route_file
+        ssl                               # :ssldir                      /etc/puppet/ssl
+
+    /opt/puppetlabs/bin                   # symlink targets of puppet related binaries, excluding ruby, openssl
+        cfacter@                          -> /opt/puppetlabs/puppet/bin/cfacter
+        facter@                           -> /opt/puppetlabs/puppet/bin/facter
+        hiera@                            -> /opt/puppetlabs/puppet/bin/hiera
+        mco@                              -> /opt/puppetlabs/puppet/bin/mco
+        mcollectived@                     -> /opt/puppetlabs/puppet/bin/mcollectived
```
kylog: $.02: users won't be running mcollectived from the command line so no need for a symlink.

See some discussion in https://github.com/puppetlabs/marionette-collective/pull/296 on this.

/cc @richardc

file_paths.md
joshcooper: k updated to libruby.so

file_paths.md
```diff
@@ -0,0 +1,123 @@
...
+        binder_config.yaml                # :binder_config
+        csr_attributes.yaml               # :csr_attributes
+        custom_trusted_oid_mapping.yaml   # :trusted_oid_mapping_file
+        device.conf                       # :deviceconfig
+        fileserver.conf                   # :fileserverconfig
+        puppet.conf                       # :config
+        routes.yaml                       # :route_file
+        ssl                               # :ssldir                      /etc/puppet/ssl
+
+    /opt/puppetlabs/bin                   # symlink targets of puppet related binaries, excluding ruby, openssl
```
kylog: I might drop ", excluding ruby, openssl" since it's not an exhaustive list of what is not symlinked.

file_paths.md
```diff
@@ -0,0 +1,123 @@
+Puppet Installation Layout
+==========================
+
+This table specifies the file paths in a Puppet installation and the corresponding settings, denoted as `:setting`. The 3.x column shows the difference of the specification vs the actual 3.x implementation.
```
kylog: Drop "actual"? That word implies a described-vs-actual comparison to me, but the comparison is 4.x/AIO vs 3.x.

file_paths.md
```diff
@@ -0,0 +1,123 @@
...
+            libruby.so
+            libssl.so
+            ruby
+                vendor_ruby               # ruby code
+                    cfacter.rb
+                    facter.rb
+                    hiera.rb
+                    mcollective.rb
+                    puppet.rb
+            virt-what
```
kylog: Is `virt-what` really under `lib`?

file_paths.md
joshcooper: done

file_paths.md
joshcooper: currently it's

```
# ls -la /opt/puppetlabs/puppet/lib/virt-what/
total 16
drwxr-xr-x. 2 root root   35 Feb 18 13:36 .
drwxr-xr-x. 6 root root 4096 Feb 18 13:36 ..
-rwxr-xr-x. 1 root root 8396 Feb 15 19:31 virt-what-cpuid-helper
```

file_paths.md
joshcooper: I was thinking of this document as being the specification (which AIO is working on implementing, though not yet completely). And we're comparing the spec against what was implemented previously. But perhaps the distinction is more confusing than necessary.

hlindberg: Since there is only one "3x implementation", the word "actual" has no more specificity - the one and only implementation is the "actual" one :-). (I was the one that put it there, I have been reading "actual" too many times in comments - it is actually [sic] almost never needed....).

file_paths.md
haus: That matches distro packages. They have that file in lib, and a binary in sbin or bin.

joshcooper: clarified virt-what in bin and lib.

file_paths.md
joshcooper: Makes sense, I removed this symlink

file_paths.md
shrug: Per the server layout proposal, server paths will look like:

/opt/puppetlabs/server/apps/puppetserver <-- Application dir. Contains FHS-like structure underneath (bin/, lib/, etc, as required by each component)
/opt/puppetlabs/server/data/puppetserver <-- Persistant data, vardir, $HOME for services that use it
/opt/puppetlabs/server/bin <-- symlinks of server binaries (from their app dirs) for convenience. Commonly used binaries (from our projects) may also link to /opt/puppetlabs/bin

joshcooper: @shrug could you provide an example of server binaries that exist in `/opt/puppetlabs/server/apps/puppetserver/bin` and are symlinked into `/opt/puppetlabs/server/bin` and `/opt/puppetlabs/bin`? I'm assuming `puppetmaster` will be?

joshcooper: > /opt/puppetlabs/server/data/puppetserver <-- Persistant data, vardir,

This means `/opt/puppetlabs/server/data/puppetserver` is the master vardir, right? As opposed to there being a `var` subdirectory in there?

shrug: Correct, yes

shrug: {quote}
@shrug could you provide an example of server binaries that exist in /opt/puppetlabs/server/apps/puppetserver/bin and are symlinked into /opt/puppetlabs/server/bin and /opt/puppetlabs/bin? I'm assuming puppetmaster will be?{quote}

Yep, puppetmaster is a good example. 

We could link *all* of the binaries to /opt/puppet/bin, but it seems like it makes sense to leave less-commonly used things (like the httpd binaries) in /opt/puppetlabs/server/bin. Things that are typically interacted with via the init system instead of directly, for example.

joshcooper:
I've gathered initial feedback from most everyone, and think this is ready to be merged. /cc @kylog @hlindberg

kylog:
The `puppetserver` binary is called `puppetmaster`? I know nothing about this, but it seems odd so I want to confirm.

/cc @shrug @cprice404

hlindberg:
I am +1 to merging

kylog:
Lgtm. Merging.

PR #24, opened 2015-02-23
(PUP-4026) Add facter's facts.d

joshcooper:
Previously we only specified the location that puppet pluginsyncs
external facts to. Unfortunately, facter's facts.d directory was not
specified.

This commit adds facter's facts.d directory to
/opt/puppetlabs/facter/facts.d. It is not located in
/opt/puppetlabs/puppet, because this facts.d directory is a facter
implementation detail, and puppet has no knowledge of it.


kylog:
:+1:

MikaelSmith:
Seems fair.

PR #25, opened 2015-02-23
(PUP-4026) Add Windows paths

joshcooper:
Adds windows paths. All paths are the same prior to AIO except for
`codedir` (new) and `vardir` is renamed from `var` to `cache` for
consistency with *nix.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -3,8 +3,15 @@ Puppet Installation Layout
 
 This table specifies the file paths in a Puppet installation and the corresponding settings, denoted as `:setting`. The 3.x column shows the difference of the specification vs the 3.x implementation.
 
-puppet-agent
-------------
+# Index
+
+* [puppet-agent (*nix)](#puppet-agent-nix)
+* [puppet-agent (Windows)](#puppet-agent-windows)
+* [puppet-db](#puppet-db)
+* [puppet-server](#puppet-server)
+* [Notes](#notes)
+
+# puppet-agent (*nix)
 
     Path                                  Setting                        3.x
     /etc/puppetlabs                                                      n/a
@@ -43,7 +50,7 @@ puppet-agent
         facts.d                           # external facts directory (not pluginsync'ed)
 
     /opt/puppetlabs/mcollective/
-        plugins
+        plugins                           # user installed plugins
 
     /opt/puppetlabs/puppet                # ruby-puppet root
         bin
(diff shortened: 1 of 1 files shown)


MikaelSmith:
:+1:

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+
+        tools
+            bin
+                elevate.exe                   # Used to elevate interactive commands
+
+
 These sections describe other Puppet packages that rely on puppet-agent to create the initial directory layout. It does not attempt to specify the full set of file paths for these packages, just cases where the other package has a dependency on puppet-agent.
 
-puppetdb
+<a id="puppet-db">puppet-db</a>
```
Iristyle: Nit - standard Markdown headers (of any h size) should work just fine here as anchors - i.e.

```
### puppet-db
```

haus:
:+1:

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+        environment.bat                       # setup LOAD_PATH
+        *_interactive.bat                     # targets for shortcuts
+
+    C:\Program Files\Puppet Labs\Puppet\cfacter
+        bin                                   # executables and dlls
+            cfacter.exe
+            libfacter.so
+            lib*.dll
+        inc                                   # cfacter headers
+            facter
```
kylog: Should we be consolidating include directories on Windows as we do on Unix (where they are consolidated under /opt/puppetlabs/include/{project1,project2}? 

Or is what's in this PR more idiomatic, i.e. to have one include per project, organized under that project?

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+directory defaults to `C:\ProgramData`. On 2003, common app data is
+under `C:\Documents and Settings\All Users\Application Data`. Also
+when installing puppet-agent 32-bit on 64-bit windows, the
+installation path defaults to `C:\Program Files (x86)\Puppet
+Labs`. The examples below assume 2008/2012 and puppet-agent (64-bit).
+
+    Path                                      Setting                        3.x
+    C:\ProgramData                                                           n/a
+
+    C:\ProgramData\PuppetLabs\code            # :codedir                     contents moved from :confdir
```
Iristyle: `contents moved from C:\ProgramData\PuppetLabs\puppet\etc (:confdir)` ...

or similar?

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+        environments                          # :environmentpath
+        hiera.yaml                            # :hiera_config
+        hieradata                             # n/a
+        modules                               # user modulepath
+
+    C:\ProgramData\PuppetLabs\mcollective\etc                                same
```
Iristyle: :+1:

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+        client.cfg
+        facts.yaml
+        server.cfg
+
+    C:\ProgramData\PuppetLabs\mcollective\var                                same
```
Iristyle: :+1:

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+        log
+            mcollective.log
+
+    C:\ProgramData\PuppetLabs\puppet\etc      # :confdir                     same
```
Iristyle: :+1:

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+        csr_attributes.yaml                   # :csr_attributes
+        custom_trusted_oid_mapping.yaml       # :trusted_oid_mapping_file
+        puppet.conf                           # :config
+        ssl                                   # :ssldir
+
+    C:\ProgramData\PuppetLabs\puppet\cache    # :vardir                      C:\ProgramData\PuppetLabs\puppet\var
```
Iristyle: :+1:

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+        facts.d                               # :pluginfactdest (pluginsync'ed)
+        lib                                   # :libdir
+        facts                                 # used to generate :factpath
+        puppet-module                         # :module_working_dir
+        reports                               # :reportdir
+        server_datadir                        # :server_data
+        state                                 # :statedir
+        yaml                                  # :yamldir
+
+    C:\ProgramData\PuppetLabs\facter                                         same
```
Iristyle: :+1: So I assume the other platforms will mirror this structure then?  (another PR forthcoming?)

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+        facts.d                               # external facts directory (not pluginsync'ed)
+
+    C:\ProgramData\PuppetLabs\mcollective
+        plugins                               # user installed plugins
+
+    C:\ProgramData\PuppetLabs\puppet\var\log  # :logdir                      same
```
Iristyle: :+1:

file_paths.md
```diff
@@ -106,16 +115,144 @@ puppet-agent
...
+        puppet.log                            # not enabled by default
+
+    C:\ProgramData\PuppetLabs\puppet\var\run  # :rundir                      same
+        agent.pid                             # :pidfile
+
+    C:\Program Files\Puppet Labs\Puppet\bin
```
Iristyle: :+1:

file_paths.md
Iristyle: Hmm... not sure.  What other native headers are there even?

haus: ruby is the only other thing we ship that has headers for windows: https://github.com/puppetlabs/puppet-win32-ruby/tree/2.1.x-x86/ruby/include/ruby-2.1.0

kylog: Same question would apply to `bin` actually. On Unix we have one bin directory; on Windows we have one per-subproject. The latter seems odd to me, but perhaps its idiomatic.

In both the `bin` and `inc` cases, either will work, but the solution we provide for Windows means a much longer path to catch all those individual directories.

Hence my question about what's idiomatic. But now that I look at the `bin` case, we've clearly been doing the one-directory-per-project approach for a while, so we might as well extend that to `inc`.

tldr: :+1:

file_paths.md
joshcooper: For facter, https://github.com/puppetlabs/facter/pull/855, and ticket filed for cfacter: https://tickets.puppetlabs.com/browse/CFACT-228

file_paths.md
joshcooper: ah good to know, will fix up

joshcooper: So getting the anchors working with spaces and `(` doesn't seem to work...

file_paths.md
joshcooper: updated

file_paths.md
joshcooper: TBH, I'm not super happy with the one-bin/lib per project approach, as for each new project, e.g. foo, we have to add `foo/bin` to `PATH` and `foo/lib` to `LOAD_PATH` in
https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L25.

We then have to add a `.bat` wrapper in `C:\Program Files\Puppet Labs\Puppet\bin\foo.bat` to call the ruby bin wrapper `C:\Program Files\Puppet Labs\Puppet\foo\bin\foo`.

That said, I think changing this would be a fair amount of work, probably not worth the trouble.

kylog: Yep, that's what I concluded too.

file_paths.md
Iristyle: I created a quick demo to show it works, but I discovered the anchors must always be in lower case - perhaps you were using upper case?

https://github.com/Iristyle/Readme-test

file_paths.md
Iristyle: Agreed.

file_paths.md
Iristyle: :+1:

Iristyle:
Aside from the trivial anchor business, :shipit:

file_paths.md
ferventcoder: Chrome has weird issues with anchors by the way.

joshcooper: @Iristyle but if I create a heading like `## puppet-agent (Windows)`, then I wasn't able to link to it. Perhaps I need to urlencode the space and parentheses? I'm inclined to punt on this.

Iristyle: You just have to remove punctuation and replace spaces with dashes...

https://github.com/Iristyle/Readme-test/blob/master/README.md#puppet-agent-windows

Iristyle: But it's not a huge deal... I just generally try to keep HTML out of Markdown if possible.

joshcooper: ah, that's easier than I thought. I'll update.

kylog:
I'm :+1: .  Are there outstanding issues?

ferventcoder:
LGTM

joshcooper:
Fixed anchors

PR #27, opened 2015-02-25
(maint) Clarify puppetserver log and pid locations

jeffmccune:
Without this patch the location of the puppetserver log files and pid files are
ambiguous.  The location inferred by the locations of the agent.pid and
puppet.log are problematic because puppetserver does not run as root and as
such will not be able to create new files in `/var/run/puppetlabs` and
`/var/log/puppetlabs`.

This pull request proposes puppetserver writable directories of
`/var/log/puppetlabs/puppetserver/` and `/var/run/puppetlabs/puppetserver/`

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -280,6 +280,13 @@ These sections describe other Puppet packages that rely on puppet-agent to creat
                 server_datadir            # :server_data
                 yaml                      # :yamldir
 
+    /var/log/puppetlabs                   # :logdir                      /var/lib/puppet/log
+        puppetserver/                     # writeable by puppetserver
+            puppetserver.log
+
+    /var/run/puppetlabs                   # :rundir                      /var/lib/puppet/run
+        puppetserver/                     # writeable by puppetserver
+            puppetserver.pid
 
 # Notes
 


haus:
CC: @shrug 

@jeffmccune this would also include heap dumps?

haus:
This proposal seems sane to me.

jeffmccune:
@haus I dunno, where do heap dumps normally get written?  I personally would expect them to go to `/var/tmp` or `/var/spool`, but that's in my pre-AIO brain.

jeffmccune:
@haus Or maybe the $HOME of the puppet user, but that'd be in /opt ...

haus:
@jeffmccune right now they go in /var/log, so it would at least be consistent with the previous behavior

haus:
@jeffmccune based on my reading of https://github.com/nwolfe/ezbake/blob/feature/master/svr363-puppet-server-ezbake-upgrade/resources/puppetlabs/lein-ezbake/template/foss/ext/redhat/init.erb#L40

shrug:
:+1:  this matches what I currently have in a working branch of ezbake

jeffmccune:
Yeah, I'm fine keeping them in /var/log right now, but I don't think we should make that part of the specification at this point in time because I'm concerned that dumping heaps to that directory will cause the filesystem to fill and logs to stop being written.

Ideally heap dumps may get lost due to a full filesystem, but logs could still keep being written.

jeffmccune:
> Yeah, I'm fine keeping them in /var/log right now

That is to say, I'm fine keeping heap dumps in the same location we keep log files, right now...  but I'd like to avoid specifying it...

haus:
@jeffmccune sounds good to me

jeffmccune:
@shrug Crap, I've been hacking on https://github.com/puppetlabs/ezbake/pull/189

How much am I duplicating your efforts?

file_paths.md
```diff
@@ -280,6 +280,13 @@ These sections describe other Puppet packages that rely on puppet-agent to creat
                 server_datadir            # :server_data
                 yaml                      # :yamldir
 
+    /var/log/puppetlabs                   # :logdir                      /var/lib/puppet/log
+        puppetserver/                     # writeable by puppetserver
+            puppetserver.log
```
joshcooper: According to the [original confluence document](https://confluence.puppetlabs.com/pages/viewpage.action?pageId=33948725), this was supposed to be

> Logs: /var/log/puppetlabs/$project.log

And PUP-3809 just said:

> /var/log/puppetlabs <= logfiles go here 

Can we make `/var/log/puppetlabs` owned by `root:puppet` and writable to the group? I assume not because `puppet-agent` creates the directory and there's likely not a `puppet` user?

Will other server-side apps run as `puppet` user? If so, will they have the same issue, e.g. puppetdb?

jeffmccune: @haus ping. ^^

haus: puppetdb runs as the puppetdb user, so it would have the same problem. Making /var/log/puppetlabs owned by root:puppet isn't feasible because the puppet-agent package doesn't make the puppet user/group.

joshcooper: So we're saying each server-side app that runs as `puppet` will need to create it's own `/var/log/puppetlabs/$project` directory? Or the `puppetserver` package is going to create a `/var/log/puppetlabs/server` directory owned by `puppet:puppet`, and other server-side apps write to?

I assume the former, because not every server will have a `puppetserver` package installed, e.g. puppetdb, console?

joshcooper: Talked this over with @haus, and sounds like the former, one subdirectory per server project. Also puppetdb runs as `puppetdb` user, not `puppet`, so that makes sense too.

:shipit:

PR #28, opened 2015-02-25
Specify account and service names

joshcooper:
Describes what accounts and services are created by the `puppetserver`
and `puppet-agent` packages, and which account each service runs as.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -13,6 +13,9 @@ This table specifies the file paths in a Puppet installation and the correspondi
 
 # puppet-agent (*nix)
 
+The package will create two services `puppet` and `mcollective`, both
+running as `root` by default. It will not create a `puppet` user or group.
+
     Path                                  Setting                        3.x
     /etc/puppetlabs                                                      n/a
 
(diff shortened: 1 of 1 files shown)


haus:
:+1:

file_paths.md
```diff
@@ -121,6 +124,10 @@ when installing puppet-agent 32-bit on 64-bit windows, the
 installation path defaults to `C:\Program Files (x86)\Puppet
 Labs`. The examples below assume 2008/2012 and puppet-agent (64-bit).
 
+The package will create two services `puppet` and `mcollective`
+running as `LocalSystem` by default. It will not create a `puppet`
+user or group.
+
```
ferventcoder: Do we need to specify what it used to be in older versions?

joshcooper: Good question. For FOSS, they are the same (I think). For PE I think everything was prefixed with `pe-`. /cc @haus ?

kylog:
Disregard my comment on "puppet-server" vs "puppetserver". I asked in the Puppet Server room and ... it was confusing. Not something we need to settle in this PR.

file_paths.md
haus: For PE, it used to be `pe-puppet` for both user/group and service name, and `pe-mcollective` for the mco service. For FOSS, the puppet user/group used to be created by the puppet package and the puppet user was used in the puppet-server and puppetmaster packages/services.

joshcooper:
@jeffmccune @camlow325 do you have any comments before merging?

camlow325:
Looks okay to me.

PR #32, opened 2015-02-26
(maint) Update file pathing to indicate what is made available via packa...

melissa:
...ging

The asterisks indicate what is installed with packaging, as opposed to what is expected where, what is added on a puppet run, what is created on a service start, etc.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -15,39 +15,40 @@ This table specifies the file paths in a Puppet installation and the correspondi
 
 The package will create two services `puppet` and `mcollective`, both
 running as `root` by default. It will not create a `puppet` user or group.
+The files annotated by an '*' indicate that they are created by package installation.
 
     Path                                  Setting                        3.x
-    /etc/puppetlabs                                                      n/a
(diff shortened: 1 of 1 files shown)


melissa:
@kylog and @joshcooper fyi, as per what we talked about earlier

file_paths.md
```diff
@@ -15,39 +15,40 @@ This table specifies the file paths in a Puppet installation and the correspondi
 
 The package will create two services `puppet` and `mcollective`, both
 running as `root` by default. It will not create a `puppet` user or group.
+The files annotated by an '*' indicate that they are created during packaging.
```
kylog: "indicate that they are created during packaging" => "are created by package installation"  ??

haus:
:+1:

file_paths.md
haus: :+1:

kylog:
:+1: 

We should probably tackle Windows/puppetserver/etc as well, but maybe another day.

PR #34, opened 2015-03-04
Add puppetmaster to the specifications

haus:
As passenger packages will be happening for the AIO/puppet 4 release
(tickets pending), this commit adds the expected layout of the
puppetmaster package. This will need updating with the expected location
of passenger within the package.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -9,6 +9,7 @@ This table specifies the file paths in a Puppet installation and the correspondi
 * [puppet-agent (Windows)](#puppet-agent-windows)
 * [puppet-db](#puppet-db)
 * [puppetserver](#puppetserver)
+* [puppetmaster](#puppetmaster)
 * [Notes](#notes)
 
 # puppet-agent (*nix)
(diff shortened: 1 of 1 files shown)


haus:
@shrug @stahnma welcome to the party

file_paths.md
```diff
@@ -305,6 +306,27 @@ The package will install a service named `puppetserver`, create a
...
             puppetserver.pid
 
+# puppetmaster
+
+The package will install a service named `puppetmaster`, create a
+`puppet` user and group, and run the service as the `puppet` user.
+
+    /opt/puppetlabs/server                # serverside apps live underneath
+        data
+            puppetmaster *                # :vardir (and $HOME for services that use it)
```
joshcooper: nit, the package will create `/opt/puppetlabs/server` and `data` too, same question below?

haus: good question. i assume it will have to. this could get awesome. for log and run, the agent package will create /var/run/puppetlabs and /var/log/puppetlabs and puppetmaster will need to depend on the agent.

haus: @joshcooper so there is an oddity here. the package will create those paths, but not own them. do we care about that distinction? (if it owned /opt/puppetlabs/server, then no other server packages could also own them. it would get weird fast.).

haus:
@joshcooper updated. as i noted, `/var/log/puppetlabs` and `/var/run/puppetlabs` will be created by the agent package, which will be a dependency of a puppetmaster package.

file_paths.md
joshcooper: I don't know enough about packaging to answer that question

haus: The * note currently indicates the package will create the path, which is still true, so I added an * for it.

joshcooper:
Could you call out in the spec that `puppetmaster` is the passenger package?

haus:
@joshcooper will do.

haus:
@joshcooper added.

PR #35, opened 2015-03-10
Remove reference to a separate service

stahnma:
Previously the specification said we would create a new service. We are
actually just going to rely upon a stock Apache installations and build
passenger within those constraints.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -319,8 +319,8 @@ The package will install a service named `puppetserver`, create a
 
 This is a compatibility package using passenger to serve a ruby based puppetmaster.
 
-The package will install a service named `puppetmaster`, create a
-`puppet` user and group, and run the service as the `puppet` user.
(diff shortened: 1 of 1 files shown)


haus:
:+1:

file_paths.md
```diff
@@ -319,8 +319,8 @@ The package will install a service named `puppetserver`, create a
 
 This is a compatibility package using passenger to serve a ruby based puppetmaster.
 
-The package will install a service named `puppetmaster`, create a
-`puppet` user and group, and run the service as the `puppet` user.
+The package will install passenger in conjunction with Apache, create a
```
joshcooper: I take that to mean the Apache install will create an `apache` service? If so, could you make that more explicit?

joshcooper: ping @stahnma

stahnma: it will have an httpd service on RedHat and Apache2 on Debian.

PR #37, opened 2015-03-13
(PUP-4194) Isolate puppet's logdir

joshcooper:
When puppet executes as root, it sets permissions for file-based
settings to either `puppet:puppet` or `root:root` depending on whether
the puppet user and group exist or not.

Previously, puppet's `logdir` setting was `/var/log/puppetlabs`. If the
puppet user didn't exist, then puppet would set permissions to
root:root with mode 0750. If you later installed `puppetserver`, it
would create its log directory in `$logdir/puppetserver`, create the
puppet user and group, start the JVM as puppet, and fail to write to
its logfile, due to permissions on the grandparent `logdir`.

This commit changes the `logdir` to `/var/log/puppetlabs/puppet`, so
that puppet can manage permissions on its `logdir` without affecting
puppetserver, or other packages.

Note mcollective.log remains where it is.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -112,9 +112,10 @@ The files annotated by an '*' indicate that they are created by package installa
             vim *
         ssl *
 
-    /var/log/puppetlabs *                 # :logdir                      /var/lib/puppet/log
-        puppet.log                        # not enabled by default
+    /var/log/puppetlabs *
         mcollective.log
+        puppet *                          # :logdir                      /var/lib/puppet/log
+            puppet.log                    # not enabled by default
 
     /var/run/puppetlabs *                 # :rundir                      /var/lib/puppet/run
         agent.pid                         # :pidfile
@@ -307,12 +308,12 @@ The package will install a service named `puppetserver`, create a
                 server_data               # :server_datadir
                 yaml                      # :yamldir
 
-    /var/log/puppetlabs                   # :logdir                      /var/lib/puppet/log
-        puppetserver/                     # writeable by puppetserver
+    /var/log/puppetlabs
+        puppetserver *                    # writeable by puppetserver
             puppetserver.log
 
     /var/run/puppetlabs                   # :rundir                      /var/lib/puppet/run
-        puppetserver/                     # writeable by puppetserver
+        puppetserver                      # writeable by puppetserver
             puppetserver.pid
 
 # puppetmaster
@@ -330,7 +331,7 @@ The package will install a service named `puppetmaster`, create a
                 server_data               # :server_datadir
                 yaml                      # :yamldir
 
-    /var/log/puppetlabs                   # :logdir                      /var/lib/puppet/log
+    /var/log/puppetlabs
         puppetmaster *                    # writeable by puppetmaster
             puppetmaster.log
 


joshcooper:
A few questions:

1. I choose `/var/log/puppetlabs/puppet` and not `puppetagent`, though perhaps it's better that the directory name match the package?
1. Should we move `/var/log/puppetlabs/mcollective.log`? If so, under the `puppet`, `puppetagent`, or `mcollective` directory? Note any change will require mcollective 2.8.2 release.
1. Should we move the puppet agent and mcollective pid files in  `/var/run/puppetlabs`? If so, same question as above, under `puppet`, `puppetagent`, `mcollective`?

/cc @richardc

kylog:
Re 1) I'd vote for `/var/log/puppetlabs/puppet` - the only known use for this directory is webrick logs, and it seems counterintuitive to me that it would be a directory named `puppetagent`. 

I also suggest that we deprecate in 4.x (and remove in 5.x) the `logdir` setting and its two consumers: `masterhttplog` (webrick only), and `puppetdlog` (not used by an OS past or present AFAICT).

Re 2) I'd incline to leaving mcollective.log be.

Re 3) I'd incline to leaving the pid files be. Having a single directory for all the pids makes more sense to me and sounds more convenient.

joshcooper:
Re 3), unfortunately we're not completely consistent about a single directory as we have:
```
/var/run/puppetlabs/{agent.pid,mcollectived.pid,master.pid}
/var/run/puppetlabs/puppetmaster/puppetmaster.pid
/var/run/puppetlabs/puppetserver/puppetserver.pid
```
But the latter two required so that puppetserver (running as puppet) can write its pid file. That said I'm fine leaving as is.

kylog:
FWIW I think of intrinsically single-file directories as a UX anti-pattern in general. E.g. as a sysadmin hunting for a pid, each nested subdirectory looks like more tab-completion keystrokes I'd just as soon save.

Not sure if others think of single-file directories the same way though :)

richardc:
Per-vendor directories for anything are also weird from a UX standpoint.  I'm trying to find logfiles/config/pidfiles for mcollective and now I have to remember that PuppetLabs made it?  I'm used to /etc/$project but /etc/$vendor/$project is somewhat rare.

joshcooper:
@richardc To make it super explicit, you're suggesting we do:

```
/var/run/puppet/agent.pid
/var/run/mcollective/mcollectived.pid
/var/run/puppetmaster/puppetmaster.pid
/var/run/puppetserver/puppetserver.pid
```

where `/var/run/{puppetmaster,puppetserver}` would have `puppet:puppet:755` permissions, `/var/run/mcollective` would have `root:root:755` permissions, and `/var/run/puppet` would either be the former if the `puppet` service user & group exist (on the master), or the latter on an agent-only node.

and 

```
/var/log/puppet/masterhttp.log         # for webrick
/var/log/mcollective/mcollective.log   # note this isn't "mcollectived.log"
/var/log/puppetmaster/puppetmaster.log
/var/log/puppetserver/puppetserver.log
```

where `/var/log/{puppetmaster,puppetserver}` would have `puppet:puppet:750` permissions, `/var/log/mcollective` would have `root:root:750` permissions, and `/var/log/puppet` would either be the former if the `puppet` service user & group exist (on the master), or the latter on an agent-only node.

/cc @kylog @haus @ahpook @johnduarte thoughts?

ahpook:
i like @richardc's suggestion better than either the top-line `puppetlabs` directory with lots of subdirectories  or the many-pid-files-in-one-directory pattern, for whatever that's worth.

kylog:
@richardc's comment referenced `/etc` as well, but it's not immediately clear to me how we'd apply it there, e.g. because we currently have `/etc/puppetlabs/code` which is not under a `project` subdirectory. 

Regardless, even just the `var` changes may be more easily hashed out in a new PR than in comments on this PR.

richardc:
@joshcooper I'm not suggesting /var/run/$application/$application.pid in all cases.  I'd prefer /var/run/$application.pid for projects that can support it (ie ones where the daemon is launched as root) and then /var/run/$application/$application.pid for tools which need a statedir created for them as part of packaging.

It was a slip to talk about /etc.  I assume that thing should have been debated to death in the early stages of the AIO standardisation and I really meant /var/run.

stahnma:
I'm trying to understand the final proposal here. 

Overall, I'd like it to feel like a single agent rather than a collection of daemons. I think keeping it under puppetlabs helps preserve that. I like the name-spaced directories for puppetlabs, as it's more the way that vendor software is managed. Also we need subdirectories in places for ownership.  

I also think it's pretty confusing for a user that puppetmaster != puppetserver in terms of logs, pids and stuff. I'm not sure what the best answer is there, but I find it confusing, and I work here. 

:memo: Also, do modern OSes even need pids? I've always thought it was weird that Puppet tried to manage its own PID and not have the init script or system do it. With systemd, you don't always write a PID. (This is probably a side bar)

joshcooper:
> Overall, I'd like it to feel like a single agent rather than a collection of daemons... Also we need subdirectories in places for ownership.

This PR continues along those lines.

> I also think it's pretty confusing for a user that puppetmaster != puppetserver in terms of logs

Yes, but it's outside the scope of this PR.

> Also, do modern OSes even need pids

Puppet and friends have behaved that way for awhile. Also something outside the scope of this PR.

stahnma:
I guess I'd still like to keep the puppetlabs namespace there in nearly every case. What's the final proposal?

joshcooper:
@stahnma final proposal is what's in this PR. It keeps the puppetlabs namespace for all the things, creating project specific directories when we have to due to permissions issues.

haus:
i'm :+1: here.

PR #40, opened 2015-04-21
(packaging) Remove reference to etc/ under app dir for server layout

shrug:
Under the current plan, server component config files will live in
/etc/puppetlabs/<component> - there will not be a config directory
under /opt/puppetlabs/server/apps/<component>.


stahnma:
:+1:

PR #46, opened 2015-07-08
Maint/master/updates

melissa:

PR #47, opened 2015-07-09
(PE-9922) Update file paths specification for File Sync service paths

rlinehan:
Add a directory `/etc/puppetlabs/code-staging` which will be the directory
where users will lay down and interact with code if they are using the file
sync service. The File Sync service will then sync this code into the codedir.

Add the File Sync service datadir at
`/opt/puppetlabs/data/puppetserver/filesync`. This is where the File Sync
storage service and client service will keep the git directories they use for
storing data.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -32,6 +32,16 @@ The files annotated by an '*' indicate that they are created by package installa
         hiera.yaml *                      # :hiera_config
         modules *                         # user modulepath
 
+    /etc/puppetlabs/code-staging          # staging directory            n/a
+        environments
+          production
+            hieradata
+            environment.conf
+            manifests
+            modules
+        hiera.yaml
+        modules
+
     /etc/puppetlabs/mcollective *
         client.cfg *
         facts.yaml *
@@ -328,6 +338,9 @@ The package will install a service named `puppetserver`, create a
         data
             puppetserver                  # :vardir (and $HOME for services that use it)
                 bucket                    # :bucketdir
+                filesync                  # file sync service datadir
+                  client
+                  storage
                 reports                   # :reportdir
                 server_data               # :server_datadir
                 yaml                      # :yamldir


rlinehan:
This PR is currently based off of #46, since that corrected the hieradata location. It should be rebased once that is merged.

rlinehan:
The file sync service will only be in PE, so it might be more appropriate to copy the current `file_paths.md` into something like `pe_file_paths.md` and then make these updates. However, from discussions it sounds like even though file sync won't be in FOSS we might still have the concept of a staging directory. It also seems like it would be really annoying and labor-intensive to have to maintain a second file that is mostly just a copy. What do other folks think?

file_paths.md
```diff
@@ -25,13 +25,23 @@ The files annotated by an '*' indicate that they are created by package installa
...
           production *
+            hieradata *
             environment.conf *
             manifests *
             modules *
         hiera.yaml *                      # :hiera_config
-        hieradata *                       # n/a
         modules *                         # user modulepath
 
+    /etc/puppetlabs/code-staging          # staging directory            n/a
```
rlinehan: Not sure whether this would be created by packaging or not.

stahnma: What's the thought behind this in /etc? Is it actually configuration? It doesn't feel like it. If feels like it belongs in a content directory under /opt to me. I apologize if I missed that part of the discussion on the thread. There was a lot going on about which implementation before we got down to details (or that's how I read it).

stahnma: (I guess we already do this with facts.d, so maybe I'm just saying too littel too late. )

stahnma:
I assume the discussion for this is mostly closed given the thread, but in Release, we're not super happy about the arrived at end. Telling the user to interact with code in a different place seems way more confusing than having the software look somewhere else (and somewhere more private than /etc).

file_paths.md
rlinehan: This would be the place where users are interacting with code. My understanding is that things that will have user interaction should usually go in /etc, and since we already have puppet code in /etc this seems like the right place for it to go?

joshcooper:
@rlinehan #46 has been merged, should be able to rebase now

rlinehan:
Rebased!

cprice404:
@stahnma can you weigh in on the thread?  I would like to make sure that your concerns are captured and brought into the conversation, but the thread is more visible than the PR comments.

rlinehan:
Ping @joshcooper @eric0 @cprice404 @stahnma for review. This has been open for a while now. Can any of y'all weigh in on this and provide feedback? Is there anyone else who should look at this?

joshcooper:
I'm :+1:, but I think we're waiting on @stahnma

rlinehan:
@stahnma any chance you could provide feedback on this soon? At this point we are proceeding with PE integration of file sync using this as a basis, and it would be good to know sooner rather than later whether we should be modifying anything we're doing.

stahnma:
Sorry all, for some reason I'm not getting notifications on this. I'll jump in on the thread.

ahpook:
I think we're good to merge.

PR #49, opened 2015-09-24
(maint) Update for facter 3

kylog:
* Remove cfacter references as it's not shipped anymore
* Update to indicate facter is now an exe

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -50,7 +50,6 @@ The files annotated by an '*' indicate that they are created by package installa
         ssl                               # :ssldir                      /etc/puppet/ssl
 
     /opt/puppetlabs/bin *                 # symlink targets of puppet related binaries
-        cfacter@ *                        -> /opt/puppetlabs/puppet/bin/cfacter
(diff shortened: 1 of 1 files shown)


file_paths.md
```diff
@@ -99,7 +97,6 @@ The files annotated by an '*' indicate that they are created by package installa
             libssl.so *
             ruby *
                 vendor_ruby *             # ruby code
-                    cfacter.rb *
                     facter.rb *
```
joshcooper: just to double check, the `facter.rb` shim exists?

ferventcoder: I would believe something exists here. A quick install could prove this?

whopper: Just verified it does in the latest puppet-agent build.

kylog: Also, FYI it is a generated file, which is why you won't find it in the facter repo. However, you would find https://github.com/puppetlabs/facter/blob/8886a003ac6459e56d78cd3d2d98c6d1e8b1898b/lib/facter.rb.in

PR #50, opened 2015-09-24
(CTH-359) Add new pxp-agent section to file_paths.md

parisiale:
Adding new section with default paths of pxp-agent for Windows and *nix
platforms.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -15,8 +15,8 @@ This table specifies the file paths in a Puppet installation and the correspondi
 
 # puppet-agent (*nix)
 
-The package will create two services `puppet` and `mcollective`, both
-running as `root` by default. It will not create a `puppet` user or group.
+The package will create the following services `puppet`, `mcollective`, and `pxp-agent`,
+all running as `root` by default. It will not create a `puppet` user or group.
 The files annotated by an '*' indicate that they are created by package installation.
 
     Path                                  Setting                        3.x
@@ -58,6 +58,11 @@ The files annotated by an '*' indicate that they are created by package installa
         puppet.conf *                     # :config
         routes.yaml                       # :route_file
         ssl                               # :ssldir                      /etc/puppet/ssl
+
+    /etc/puppetlabs/pxp-agent *
+        modules *                         # stores configuration files for pxp-agent modules
+            pxp-module-puppet.conf *      # configuration file of the pxp module puppet
+        pxp-agent.conf                    # pxp-agent configuration file
 
     /opt/puppetlabs/bin *                 # symlink targets of puppet related binaries
         facter@ *                         -> /opt/puppetlabs/puppet/bin/facter
@@ -121,14 +126,20 @@ The files annotated by an '*' indicate that they are created by package installa
         ssl *
         VERSION                           # puppet-agent package version
 
+    /opt/puppetlabs/pxp-agent *
+        spool                             # directory containing results of pxp-agent modules
+
     /var/log/puppetlabs *
         mcollective.log
         puppet *                          # :logdir                      /var/lib/puppet/log
             puppet.log                    # not enabled by default
+        pxp-agent
+            pxp-agent.log                 # enabled by default
 
     /var/run/puppetlabs *                 # :rundir                      /var/lib/puppet/run
         agent.pid                         # :pidfile
         mcollectived.pid
+        pxp-agent.pid
 
 # puppet-agent (windows)
 
@@ -140,9 +151,9 @@ when installing puppet-agent 32-bit on 64-bit windows, the
 installation path defaults to `C:\Program Files (x86)\Puppet
 Labs`. The examples below assume 2008/2012 and puppet-agent (64-bit).
 
-The package will create two services `puppet` and `mcollective`
-running as `LocalSystem` by default. It will not create a `puppet`
-user or group.
+The package will create the following services `puppet`, `mcollective`,
+and `pxp-agent`, all running as `LocalSystem` by default. It will not
+create a `puppet` user or group.
 
     Path                                      Setting                        3.x
     C:\ProgramData                                                           n/a
@@ -171,7 +182,7 @@ user or group.
                 mcollective.log
 
     C:\ProgramData\PuppetLabs\puppet
-    -------------------------------------------------------------------------------------------------------               
+    -------------------------------------------------------------------------------------------------------
         cache                                 # :vardir                      C:\ProgramData\PuppetLabs\puppet\var
             bucket                            # :bucketdir
             client_yaml                       # :clientyamldir
@@ -185,7 +196,7 @@ user or group.
             server_data                       # :server_datadir
             state                             # :statedir
             yaml                              # :yamldir
-    -------------------------------------------------------------------------------------------------------               
+    -------------------------------------------------------------------------------------------------------
         etc *                                 # :confdir                     same
             auth.conf                         # :rest_authconfig
             autosign.conf                     # :autosign
@@ -197,12 +208,24 @@ user or group.
             puppet.conf *                     # :config
             routes.yaml                       # :route_file
             ssl                               # :ssldir
-    -------------------------------------------------------------------------------------------------------               
+    -------------------------------------------------------------------------------------------------------
         var *
             log                               # :logdir                      same
                 puppet.log                    # not enabled by default
             run                               # :rundir                      same
                 agent.pid                     # :pidfile
+
+    C:\ProgramData\PuppetLabs\pxp-agent *
+        etc *
+            pxp-agent.conf                    # pxp-agent configuration file
+            modules *                         # stores configuration files for pxp-agent modules
+                pxp-module-puppet.conf *      # configuration file of the pxp module puppet
+        var *
+            log
+                pxp-agent.log                 # enabled by default
+            spool                             # directory containing results of pxp-agent modules
+            run
+                pxp-agent.pid
 
     C:\Program Files\Puppet Labs\Puppet
         VERSION                               # puppet-agent package version
@@ -225,7 +248,7 @@ user or group.
         inc *                                 # facter headers
             facter *
         lib *
-            facter.rb *                      # ruby bindings
+            facter.rb *                       # ruby bindings
 
     C:\Program Files\Puppet Labs\Puppet\hiera *
         bin *


file_paths.md
```diff
@@ -339,6 +340,39 @@ The package will install a service named `puppetmaster`, create a
...
+            puppet.cfg *                  # configuration file of the `puppet` module
+
+## Windows
+
+    C:\ProgramData\PuppetLabs\pxp-agent *
+        var *
+            spool                         # :spooldir
+            modules *                     # :modulesdir
+                puppet *                  # executes Puppet and provides results of Puppet runs
+            logs *                        # :logdir
```
ferventcoder: Casing and plural - just `log`.

file_paths.md
```diff
@@ -339,6 +340,39 @@ The package will install a service named `puppetmaster`, create a
...
+This package executes Puppet and replies to PXP requests.
+
+## *nix
+
+    /opt/puppetlabs/pxp-agent *
+        spool                             # :spooldir
+        modules *                         # :modulesdir
+            puppet *                      # executes Puppet and provides results of Puppet runs
+
+    /var/log/puppetlabs/pxp-agent *       # :logdir
```
ferventcoder: Is it really uppercase on POSIX?

file_paths.md
```diff
@@ -339,6 +340,39 @@ The package will install a service named `puppetmaster`, create a
...
+
+## Windows
+
+    C:\ProgramData\PuppetLabs\pxp-agent *
+        var *
+            spool                         # :spooldir
+            modules *                     # :modulesdir
+                puppet *                  # executes Puppet and provides results of Puppet runs
+            logs *                        # :logdir
+                pxp-agent.log             # enabled by default
```
ferventcoder: Why the uppercase for the file extension?

ferventcoder:
I don't see where the actual binaries will drop in this PR. 

https://github.com/puppetlabs/puppet-specifications/blame/master/file_paths.md#L62-L67 (sorry for using blame, apparently Github doesn't have a way of looking at markdown lines)

https://github.com/puppetlabs/puppet-specifications/blame/master/file_paths.md#L213-L233

file_paths.md
ferventcoder: Same on line 355.

file_paths.md
ferventcoder: Nevermind, the font was telling me this was uppercase. Ugh. Sorry. Nothing to see here.

file_paths.md
ferventcoder: Nevermind, the font made this look uppercase to me. Ugh. Sorry. Nothing to see here.

file_paths.md
ferventcoder: Nevermind, the font made this look uppercase to me.

parisiale:
We decided to deal with binary paths in a separate ticket.

file_paths.md
```diff
@@ -339,6 +340,39 @@ The package will install a service named `puppetmaster`, create a
...
+            spool                         # :spooldir
+            modules *                     # :modulesdir
+                puppet.bat *              # executes Puppet and provides results of Puppet runs
+            log *                         # :logdir
+                pxp-agent.log             # enabled by default
+        etc *
+            pxp-agent.cfg                 # :configfile
+            modules.d *                   # :modules_configdir
+                puppet.cfg *              # configuration file of the `puppet` module
+
```
joshcooper: Thanks @parisiale for putting this together. Would you mind moving these sections into `puppet-agent` so that it's clear how the pxp-agent component is laying out files relative to other components, e.g. puppet, mcollective, etc

stahnma:
@shrug keep an eye on this one.

file_paths.md
```diff
@@ -339,6 +340,39 @@ The package will install a service named `puppetmaster`, create a
...
             puppetmaster.pid
 
+# pxp-agent
+
+This package executes Puppet and replies to PXP requests.
+
+## *nix
+
+    /opt/puppetlabs/pxp-agent *
+        spool                             # :spooldir
```
joshcooper: Are `:spooldir`, `:modulesdir`, etc puppet settings? or pxp-agent settings? We originally listed puppet settings in this document so it was clear how to obtain those values in an acceptance test using `puppet agent --configprint vardir` for example. If those are pxp-agent settings I'm not sure how to disambiguate so people don't think they are puppet settings.

file_paths.md
```diff
@@ -339,6 +340,39 @@ The package will install a service named `puppetmaster`, create a
         puppetmaster *                    # writeable by puppetmaster
             puppetmaster.pid
 
+# pxp-agent
+
+This package executes Puppet and replies to PXP requests.
```
joshcooper: Could you add the name of the pxp-agent service on *nix and windows, e.g. `pxp-agent` or `pxp-agentd` to the section we talk about other services?

file_paths.md
parisiale: Will do.

file_paths.md
parisiale: They are pxp-agent configuration options.
I'll remove them to avoid confusion.

file_paths.md
parisiale: Will do.

file_paths.md
```diff
@@ -15,8 +15,8 @@ This table specifies the file paths in a Puppet installation and the correspondi
 
 # puppet-agent (*nix)
 
-The package will create two services `puppet` and `mcollective`, both
-running as `root` by default. It will not create a `puppet` user or group.
+The package will create three services `puppet`, `mcollective`, and `pxp-agent`,
```
ferventcoder: I'd just change `three` to `the following`

file_paths.md
```diff
@@ -140,9 +153,9 @@ when installing puppet-agent 32-bit on 64-bit windows, the
 installation path defaults to `C:\Program Files (x86)\Puppet
 Labs`. The examples below assume 2008/2012 and puppet-agent (64-bit).
 
-The package will create two services `puppet` and `mcollective`
-running as `LocalSystem` by default. It will not create a `puppet`
-user or group.
+The package will create three services `puppet`, `mcollective`, and
```
ferventcoder: same comment as above.

file_paths.md
```diff
@@ -204,6 +217,20 @@ user or group.
...
 
+    C:\ProgramData\PuppetLabs\pxp-agent *
+        etc *
+            pxp-agent.cfg                     # pxp-agent configuration file
+            modules.d *                       # directory where pxp-agent modules are stored
+                puppet.cfg *                  # ../../var/modules/puppet configuration
+        var *
+            log
+                pxp-agent.log                 # enabled by default
+            modules *
```
Iristyle: This `modules` binary directory doesn't seem correct to me given it's inside of a user's data directory.

On *nix, we have a dir of `/etc/puppetlabs/pxp-agent/modules.d` for stored modules, and `/opt/puppetlabs/pxp-agent/modules/` for their binaries.

But on Windows we're proposing everything for Windows be placed under the Windows equivalent of `/etc/puppetlabs/pxp-agent` (`C:\ProgramData\PuppetLabs\pxp-agent`).  I assume that `modules.d` is separate from `modules` for a reason, so shouldn't we preserve that on Windows?

I don't think MCO is quite a good parallel to look at either, because I don't think MCO separated plugins from their binaries?  At least in the spec its `/opt/puppetlabs/mcollective/plugins` vs `C:\ProgramData\PuppetLabs\mcollective\plugins` (which is a little weird).

file_paths.md
```diff
@@ -59,6 +59,11 @@ The files annotated by an '*' indicate that they are created by package installa
         routes.yaml                       # :route_file
         ssl                               # :ssldir                      /etc/puppet/ssl
 
+    /etc/puppetlabs/pxp-agent *
+        modules.d *                       # directory where pxp-agent modules are stored
+            puppet.cfg *                  # /opt/puppetlabs/pxp-agent/modules/puppet configuration
+        pxp-agent.cfg                     # pxp-agent configuration file
```
kylog: I suggest we use the extension `.conf` going forward rather than `.cfg`.

The former is consistent with:
* puppet
* puppetdb
* puppetserver
* misc new PE tools coming out soon like `puppet-job` and `puppet-access`

Thoughts?

file_paths.md
```diff
@@ -121,14 +126,22 @@ The files annotated by an '*' indicate that they are created by package installa
         ssl *
         VERSION                           # puppet-agent package version
 
+    /opt/puppetlabs/pxp-agent *
+        modules *
+            puppet *                      # executes Puppet and provides results of Puppet runs
```
kylog: Should we name this `puppet` or `pxp-module-puppet`? I have a mild leaning toward the latter, even though it has redundant elements (since we're in a pxp-agent modules directory), because it matches the repository name and makes it easier to `find` the module when one forgets where it landed.

file_paths.md
```diff
@@ -204,6 +217,20 @@ user or group.
             run                               # :rundir                      same
                 agent.pid                     # :pidfile
 
+    C:\ProgramData\PuppetLabs\pxp-agent *
+        etc *
+            pxp-agent.cfg                     # pxp-agent configuration file
+            modules.d *                       # directory where pxp-agent modules are stored
+                puppet.cfg *                  # ../../var/modules/puppet configuration
```
kylog: Same comment here about `.conf` vs `.cfg`.

file_paths.md
kylog: Does the module need its own config file? If it's just for specifying the path to puppet, perhaps that could be hardwired in the code (and optionally overridden on the CLI for development)?

file_paths.md
```diff
@@ -204,6 +217,20 @@ user or group.
...
+        var *
+            log
+                pxp-agent.log                 # enabled by default
+            modules *
+                puppet.bat *                  # executes Puppet and provides results of Puppet runs
```
kylog: I'm fuzzy on this but does this need to be a batch file on Windows? Or can the pxp-agent simply run the ruby script directly?

file_paths.md
```diff
@@ -59,6 +59,11 @@ The files annotated by an '*' indicate that they are created by package installa
         routes.yaml                       # :route_file
         ssl                               # :ssldir                      /etc/puppet/ssl
 
+    /etc/puppetlabs/pxp-agent *
+        modules.d *                       # directory where pxp-agent modules are stored
```
kylog: Why `modules.d` vs `modules`?

file_paths.md
parisiale: I changed that after looking at the fixes done by @MikaelSmith to get pxp-agent unit tests working on Windows: https://github.com/puppetlabs/pxp-agent/tree/master/lib/tests/resources/modules.

I'm not sure if this is necessary. @ferventcoder @Iristyle, can you help me here?

file_paths.md
parisiale: Wrong comment by the way.
Since we use different directory trees, I agree in using `modules` for both.

file_paths.md
parisiale: Agree on using `.conf`.

In general, each module config file provides configuration parameters used by its actions. @ploubser can you expand on that?

file_paths.md
parisiale: Agree.

file_paths.md
parisiale: `modules.d` contains configuration files whereas `modules` the actual scripts. 
@kylog, below, suggests to use the same name for both directories.

My intent was to let scripts and config files to reside on different directory trees, since we have different locations. But I have no strong opinion.

Should scripts and configuration files reside in the same location?
What are the appropriate location(s)?

file_paths.md
ploubser: Currently the only configuration value is the path to puppet. If this was to remain the case then I guess we could hardcode it.

I do however feel that leaving the config file around might be the better plan in the long term. Right now the module fails very quietly and I can see value in adding a debug or verbose config file option that will allow us to fail louder, possibly making the module easier to debug for support in the wild.

file_paths.md
ploubser: This renaming will require some ticketed work.

file_paths.md
ferventcoder: It really depends. If you want to call it directly outside of a service, the .bat file is absolutely necessary.

ferventcoder: After looking at where this is, I'm not sure what it's purpose is at all. :/

ferventcoder: If it is calling Puppet, it should use the file provided in `C:\Program Files\Puppet Labs\Puppet\bin\puppet.bat` (or `C:\Program Files (x86)\Puppet Labs\Puppet\bin\puppet.bat`) unless there is a really good reason to duplicate.

ferventcoder: That or as suggested by @kylog, just execute the ruby script directly.

ferventcoder: Perhaps it would be better to first understand the purpose of what you want to do to be able to suggest the most relevant course for addressing it on Windows. @parisiale :)

file_paths.md
Iristyle: Configuration data should definitely live under `C:\ProgramData\PuppetLabs`.

If scripts are a write-once sort of thing, it could make sense to put them in a global `\Program Files\` rooted directory.  But on the flip side it could be easier to manage a single path if they're all lumped in together (like MCO).  I also don't have a really strong opinion on that aspect, but the more secure option is to place executables under `\Program Files\`.

My only strong opinion here is that Windows mirror the same intent as *nix.

file_paths.md
kylog: Yep, that makes sense. So yeah, keep the two config files. 

I'd suggest the module one here should be called `pxp-module-puppet.conf`.

file_paths.md
kylog: @ploubser that sounds ominous. I assume anything here that's a rename will require *some* work. Are you saying this one is notably or prohibitively sticky?

file_paths.md
kylog: @ferventcoder this script is some ruby glue that allows `pxp-agent` to ask questions of puppet: https://github.com/puppetlabs/pxp-module-puppet/blob/master/modules/puppet

And yeah @parisiale, I don't know why Michael had a `.bat` wrapper. Do we *need* a particular file extension in order to be able to execute something?

Anyway, this may be more minutia than we need to definitively sort out in *this* PR. I.e. if we sort out the directory structure now, and later change our mind to make this not be a `.bat` file, we can just come back and update this.

file_paths.md
kylog: Yeah this is config for the module not the module binary itself. So given that, I think that this is in the right place.

What may have made this confusing is that this PR includes both binary and config file paths for *nix, but only the config file paths for windows.

file_paths.md
ploubser: @kylog This one isn't prohibitively sticky, but it does require renaming and testing in pxp-puppet and the mock agent. Not a big deal, just something we need to be aware of.

file_paths.md
parisiale: Thanks @ferventcoder .
Will avoid adding this to the specs and leave just the directory, as suggested by @kylog.

file_paths.md
parisiale: I will leave `C:\ProgramData\PuppetLabs\pxp-agent\etc\modules` as the directory for module config files.

@kylog , the PR is currently specifying `C:\ProgramData\PuppetLabs\pxp-agent\var\modules` as the directory for module binaries on Windows. I will follow @Iristyle 's suggestion and put them under `\Program Files\'.

Do we agree on `C:\Program Files\Puppet Labs\Puppet\pxp-agent\modules`?

file_paths.md
parisiale: Agree.

kylog:
Seems like we're getting close-ish. I think @joshcooper wanted to review this also.

Once we are done, @parisiale can you squash these all down to a single commit?

file_paths.md
```diff
@@ -197,13 +210,25 @@ user or group.
...
+    -------------------------------------------------------------------------------------------------------
         var *
             log                               # :logdir                      same
                 puppet.log                    # not enabled by default
             run                               # :rundir                      same
                 agent.pid                     # :pidfile
 
+    C:\ProgramData\PuppetLabs\pxp-agent *
+        etc *
+            pxp-agent.conf                    # pxp-agent configuration file
```
Iristyle: The question has come up before with respect to MCO and shipping a "default" config.  What is the approach with `pxp-agent`? See [MCO-727](https://tickets.puppetlabs.com/browse/MCO-727)

file_paths.md
```diff
@@ -197,13 +210,25 @@ user or group.
...
+            modules *                         # stores configuration files for pxp-agent modules
```
Iristyle: I think modules should be created by packaging (hence the `*`), but I don't think that `pxp-module-puppet` is part of packaging, and won't come down until a connection is established with the master, so I believe we want to remove the `*` from `pxp-module-puppet.conf`

file_paths.md
```diff
@@ -252,6 +277,10 @@ user or group.
         lib *
             puppet.rb *
 
+    C:\Program Files\Puppet Labs\Puppet\pxp-agent *
+        modules *                             # contains pxp-agent modules (ruby binaries)
+            pxp-module-puppet *               # executes Puppet and provides results of Puppet runs
```
Iristyle: Same comment above about `pxp-module-puppet` removing the `*`.  Packaging should create the parent `modules` directory, but I don't think `pxp-module-puppet` comes down until after the puppet agent first runs and configures the agent node, right?

file_paths.md
kylog: I was assuming it was part of packaging. Moar email needed!! :)

file_paths.md
kylog: Same comment here.

file_paths.md
joshcooper: I'd expect pxp-agent to be able to execute `C:\Program Files\Puppet Labs\puppet\bin\puppet.bat` directly.

I think @MikaelSmith unit tests were that way, because on windows you can only execute a ruby script, e.g. `.\foo.rb` if there is a ruby application registered to handle files with that extension, which shows up in the `PATHEXT` environment variable. So he created a bat wrapper script to execute ruby with the script as an argument. Ruby itself does similar things with `gem.bat`, `irb.bat`, etc

Also the `%~dp0` magic is to get the directory that the .bat file is currently in, so that you can build a relative path and have it work if the files are relocated, e.g. user decides to install somewhere else.

Going back to the pxp-agent executing puppet question, I'm not sure you'll know where puppet is installed exactly, e.g. `C:\Program Files` vs `C:\Program Files (x86)` vs `D:\User\Specified\Path`. You could use a relative path based on the location of the pxp-agent binary (facter does something similar to locate ruby's dll/so to evaluate ruby-based facts, and that functionality is provided in leatherman) or assume `puppet` is in the PATH (and ensure the service is configured correctly).

joshcooper:
:+1: once packaging questions are resolved

file_paths.md
kylog: FYI, there's a PR for managing pxp-agent and friends at https://github.com/puppetlabs/puppetlabs-puppet_enterprise/pull/602.

file_paths.md
```diff
@@ -121,14 +126,22 @@ The files annotated by an '*' indicate that they are created by package installa
         ssl *
         VERSION                           # puppet-agent package version
 
+    /opt/puppetlabs/pxp-agent *
+        modules *                         # contains pxp-agent modules (ruby binaries)
+            pxp-module-puppet *           # executes Puppet and provides results of Puppet runs
```
shrug: Same comment as below - is this installed by packaging or does it come down after the puppet-agent run?

file_paths.md
```diff
@@ -59,6 +59,11 @@ The files annotated by an '*' indicate that they are created by package installa
         routes.yaml                       # :route_file
         ssl                               # :ssldir                      /etc/puppet/ssl
 
+    /etc/puppetlabs/pxp-agent *
+        modules *                         # stores configuration files for pxp-agent modules
```
shrug: I'm not fond of calling this 'modules', but it looks like that's already been hashed out in the ticket. I feel like this is likely to confuse people, however.

file_paths.md
Iristyle: Just to be clear, `pxp-module-puppet` here is https://github.com/puppetlabs/pxp-module-puppet/blob/master/modules/puppet ? 

Double-checking since the filename is different.

file_paths.md
Iristyle: I think we have enough buy-in on the email thread for this to be a part of packaging, so my prior comment here is invalidated.

Now on to determining if we're adding a submodule to `pxp-agent` or collapsing repos.

file_paths.md
kylog: Yes that's the file. Once this PR is agreed on and merged, we'll start updating code where there are mismatches.

kylog:
Wow, this PR is getting challenging to navigate! I may be groggy but I couldn't find where to respond inline so I'll just quote here ...

@joshcooper wrote:

> I think I understand about the extra puppet.bat now... So pxp-agent needs to run pxp-module-puppet, and that is a ruby script, so can't be executed directly on windows (with a posix shebang line). You will probably need to include a .bat wrapper like @MikaelSmith had, and have that run %~dp0ruby.exe puppet where the latter is the puppet ruby script in pxp-module-puppet, and not puppet.bat or puppet.rb in puppet.

> If you go this route, then all ruby-based pxp-modules will need to have .bat wrappers on windows. Is this what we want? Or should modules have .rb extensions, and modify pxp-agent to always invoke ruby puppet.rb on windows, and for non-windows, execute puppet.rb directly and rely on the shebang line to resolve ruby?

Now that we understand the why of the `.bat` file, I think that's my preference. It seems like poor separation of concerns for pxp-agent to know what `.rb` files means and do something special with that, only on windows. 

Whereas it would be super clear what a `.bat` file is, and then pxp-agent can be a smidge simpler.

That said, my leaning here is mild.

At least we know enough now, to answer that one way or the other in this spec. :)

file_paths.md
parisiale: Without the PCP configuration (server WebSocket URL, cert paths), pxp-agent can't start. The required parameters are specified in the [README](https://github.com/puppetlabs/pxp-agent). 

We discussed about removing the config file requirement introduced in [CTH-345](https://tickets.puppetlabs.com/browse/CTH-345) in favour of the previous behaviour (exit if the entire PCP configuration was not provided).

file_paths.md
parisiale: I don't have a strong opinion. 
We've discussed about naming 'modules' several times. Last time [here](https://tickets.puppetlabs.com/browse/CTH-359?focusedCommentId=218679&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-218679) :)

file_paths.md
parisiale: I left it without the asterisk in my last commit.
We made [CTH-380](https://tickets.puppetlabs.com/browse/CTH-380) to exhume the previous behaviour.

parisiale:
@joshcooper @kylog , added the .bat path for the pxp module puppet:
'''
C:\Program Files\Puppet Labs\Puppet\pxp-agent\modules\pxp-module-puppet.bat
'''

parisiale:
@Iristyle @kylog @shrug , I left asterisks for pxp module puppet script and conf (up to packaging).

Do we agree on that? Do we need to specify submodules?

parisiale:
@Iristyle leaving pxp-agent.conf without asterisk; we have a ticket for not requiring such file

file_paths.md
ferventcoder: From the location where it is, it's hard to see how the actual module would get there from a puppet sync. Unless I'm missing something?

ferventcoder:
> Now that we understand the why of the .bat file, I think that's my preference. It seems like poor separation of concerns for pxp-agent to know what .rb files means and do something special with that, only on windows.

@joshcooper @kylog I'm actually more of a fan of not requiring an extra .bat file. It's an extra step that may be oft overlooked by developers. I'd prefer enhancing the agent to know what to do in cases where the platform requires something just a little different. 

I'm not sure I agree with the point on poor separation of concerns. A cross-platform tool should know about the nuances of every platform and deal with them in the way that has the least impact on developers.

joshcooper:
If we expect users to always write pxp modules in ruby, then it seems like a win to not require .bat files. But I'm assuming there will only be one module to support in the immediate future, so perhaps we look at that optimization in the future?

parisiale:
Getting rid of the .bat file and executing external modules in a platform-specific way is a trivial change for pxp-agent.
If we think that it's the way to go, I suggest to do it now.

kylog:
This is the only planned module and there's no roadmap currently for other PL modules or user-provided modules. We don't want to rule that out though.

If it's trivial to ditch the .bat files and we prefer that idea, yes, let's go for it now.

parisiale:
Removed the pxp-module-puppet script paths, as external modules will be part of pxp-agent.
Modules config paths will remain as previously determined.

Created [CTH-386](https://tickets.puppetlabs.com/browse/CTH-386) to execute external modules in Windows without .bat scripts.

kylog:
Last call, people!

:+1: here (but do please squash to a single commit).

joshcooper:
:+1:

parisiale:
Squashed!

kylog:
rejoicing

ferventcoder:
@kylog wow.

PR #56, opened 2015-10-20
(PCP-101) Annotate pxp-agent's log and spool directories

geoffnichols:
This commit annotates pxp-agent's log and spool directories
to indicate they are created by package installation.

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -227,8 +227,6 @@ create a `puppet` user or group.
             log *
                 pxp-agent.log                 # enabled by default
             spool *                            # directory containing results of pxp-agent modules
-            run *
-                pxp-agent.pid
 
     C:\Program Files\Puppet Labs\Puppet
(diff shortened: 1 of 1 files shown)


ferventcoder:
LGTM, do the Windows paths need updated as well?

geoffnichols:
Updated commit to add Windows paths from the linked PR.

ferventcoder:
Thanks @geoffnichols !

I'm +1 on this.

ferventcoder:
@kylog @parisiale @joshcooper Will let you verify and merge.

parisiale:
:+1: 
Note that part of PCP-17, we decided to not write the PID to file on Windows, so `C:\ProgramData\PuppetLabs\pxp-agent\var\run` is not used currently and can be removed from the specs.

geoffnichols:
Added commit to remove the pxp-agent PID file from the Windows specs.

PR #57, opened 2015-10-22
(PA-24) File Path specifications changes for PXP-Agent

jcoconnor:
Update to the file_path.md file to capture recent pxp-agent changes

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -222,11 +222,12 @@ create a `puppet` user or group.
         etc *
             pxp-agent.conf                    # pxp-agent configuration file
             modules *                         # stores configuration files for pxp-agent modules
-                pxp-module-puppet.conf        # configuration file of the pxp module puppet
+                pxp-module-puppet.conf        # configuration file of the pxp module puppet (optionally, to override puppet.bat location)
         var *
             log *
                 pxp-agent.log                 # enabled by default
-            spool *                            # directory containing results of pxp-agent modules
+            spool *                           # directory containing results of pxp-agent modules
+            run *
 
     C:\Program Files\Puppet Labs\Puppet
         VERSION                               # puppet-agent package version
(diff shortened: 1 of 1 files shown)


jcoconnor:
@Iristyle Looks like we ship the Unit Tests .exe. (e.g. cpp-pcp-client-unittests.exe) - I checked on facter and these are not listed in the spec - is this ok ?

Would a more general ticket be worthwhile to trawl through all the directories to make sure the spec is up to date ?

richardc:
We probably shouldn't ship unittests.exe, they're not intended to work outside of a checkout in a development/build cycle.  I'd hope it's just a cmake change to mark it as something to build and not install.

file_paths.md
```diff
@@ -222,13 +222,12 @@ create a `puppet` user or group.
         etc *
             pxp-agent.conf                    # pxp-agent configuration file
             modules *                         # stores configuration files for pxp-agent modules
-                pxp-module-puppet.conf        # configuration file of the pxp module puppet
+                pxp-module-puppet.conf        # configuration file of the pxp module puppet (if required)
```
Iristyle: Can probably change `(if required)` to `(optional)` or `(optionally, to override puppet.bat location)`

file_paths.md
```diff
@@ -281,8 +280,10 @@ create a `puppet` user or group.
     C:\Program Files\Puppet Labs\Puppet\pxp-agent *
         bin *
             pxp-agent.exe *
+            lib*.dll *
```
Iristyle: Might want to note what these are for in the comments

```
lib*.dll *                    # gcc, mingw, boost libraries
```

Could we also add `libeay32.dll` to the OpenSSL section with `ssleay32.dll`?

jcoconnor:
@Iristyle Have just updated the lib*.dll comments for both pxp-agent and puppet-agent.

What do you mean "Could we also add libeay32.dll to the OpenSSL section with ssleay32.dll?" ?
Do you mean fixing the files/comments for "C:\Program Files\Puppet Labs\Puppet\sys" ?

Iristyle:
@jcoconnor yes, I mean adding to the section for `C:\Program Files\Puppet Labs\Puppet\sys`.  pxp-agent also attempts to load `libeay32.dll`, so I figured it was worth mentioning there as well.

richardc:
:+1: this looks sane to me.

PR #58, opened 2015-12-02
(PE-12458) File path changes for PE client tools

jdwelch:
Update to the file_path.md file to capture recent additions for PE client tools

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -165,7 +165,7 @@ create a `puppet` user or group.
     Path                                      Setting                        3.x
     C:\ProgramData                                                           n/a
 
-    C:\ProgramData\PuppetLabs/client-tools    # default client tool settings
+    C:\ProgramData\PuppetLabs\client-tools    # default client tool settings
         puppet-access.conf *
         puppet-orchestrator.conf *
 
--- a/file_paths.md
+++ b/file_paths.md
@@ -22,6 +22,10 @@ The files annotated by an '*' indicate that they are created by package installa
     Path                                  Setting                        3.x
     /etc/puppetlabs *                                                    n/a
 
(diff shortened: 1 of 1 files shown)


file_paths.md
```diff
@@ -161,6 +165,11 @@ create a `puppet` user or group.
     Path                                      Setting                        3.x
     C:\ProgramData                                                           n/a
 
+    C:\ProgramData\PuppetLabs/client-tools    # default client tool settings
```
shrug: There's a '/' where there should be a '\'

Do we need a config dir here? like the "C:\ProgramData\PuppetLabs\puppet\etc" below? I'm not sure what else might land in the C:\ProgramData\PuppetLabs\client-tools directory

jdwelch: Ah, i'll fix the copypasta error. I'm not sure what else would end up there, but presuming there is another tool every release, and we don't do the work to consolidate all client configs into one file (which would be the goal, but here we are), there would be the 1:1 correspondence between tool and config file.

MosesMendoza:
+1

jdwelch:
Can one of @shrug @MosesMendoza merge this if you're happy with it?

shrug:
ping @joshcooper @kylog Do y'all have any input here? I'll merge otherwise

kylog:
:+1:

joshcooper:
LGTM. One side note, the directory `C:\ProgramData\PuppetLabs\client-tools` is writable by all users, and the MSI should restrict permissions on that directory as it will presumably contain passwords, etc, e.g. https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/puppet.wxs#L296-L301. It's not hard to do, but we need to make sure that happens. /cc @Iristyle

PR #59, opened 2015-12-04
(PUP-5561) Clarify rules for autoloading puppet functions

hlindberg:
This adds language to the puppet-functions.md that clarifies the rules for naming puppet functions and how autoloading and ad-hoc manifest defined functions work.

### DIFF
--- a/language/puppet-functions.md
+++ b/language/puppet-functions.md
@@ -51,29 +51,98 @@ possible to call a puppet function with a lambda, and possibly also type a param
 
 Function Definition - Scope & Autoloading
 ---
-Puppet Functions are autoloaded from modules. They are located under <module-root>/functions.
-The filename must match the simple (non namespaced name part) of the function. Thus, a function 'min' in module 'testmodule' is placed like this:
+
+### Autoloading
+
+Namespaced Puppet Functions are auto loaded from modules and the environment when located under <module-root>/functions or <environment-root>/functions respectively. 
+
+Auto loaded puppet functions are always namespaced; in a module using the module name, and in an environment by using the special name `environment` (i.e. not the *name* of the environment since that typically changes as code is being developed, tested, put into production and then maintained, etc.).
+
+The name of the .pp file must match the simple (non namespaced name part) of the function. Thus, a function **'testmodule::min'** in module 'testmodule' is located like this:
 
     testmodule
       |- functions
          | min.pp
-         
-The functions defined this way in modules must be name-spaced. Thus the contents of this `min.pp` is:
+
+With the following contents in `min.pp` (note use of full namespace):
 
     function testmodule::min($a, $b) {
       # ...
     }
 
-It is allowed to defined Puppet Functions in the environment's logic (i.e. `site.pp`, or in the manifests loaded from the manifests directory). It is also allowed to defined functions directly
-on the command line when running puppet apply.
+And a function **'environment::min'** in a 'production' environment like this:
+
+    production
+      |- functions
+         | min.pp
+         
+With the following contents in `min.pp` (note use of namespace 'environment'):
+
+    function environment::min($a, $b) {
+      # ...
+    }
+
+Nested name spaces are allowed in both modules and the environment - e.g. a function **'testmodule::math::min'** would be located like this:
+
+    testmodule
+      |- functions
+         |- math
+            | min.pp
+
+With the following contents in `min.pp` (note use of full namespace):
+
+    function testmodule::math::min($a, $b) {
+      # ...
+    }
 
 Rules:
 
-* autoloaded functions in modules must have qualified names
-* functions that are defined in modules, but that are not autoloaded (loaded as part of something   
-  else) should also be defined in the module's namespace (it should be an error if an attempt is 
-  made to name it differently).
-* functions defined in the environment may have any name (global or namespaced)
-* A function loaded in the environment shadows functions from modules, but not system functions
+* Loading is performed by mapping the fully qualified function name to a 4.x Ruby function path, and a puppet function path, then:
+  * if the ruby path exist this 4.x Ruby function is loaded (and search stops)
+  * if the puppet path exists this Puppet function is loaded (and search stops)
+  * last, an attempt is made to load a 3.x Ruby function
+* Files containing an auto loaded function may only contain a single function (or an error is raised and evaluation stops).
 
+> Note:
+> 
+> Function loading only searches a set of distinct paths based on the fully qualified name of
+> the function.
+> Contrast this with other kinds of automatic loading of classes and user defined resource types
+> where a search is made using a widening of the namespace, and finally reaching
+> a modules `manifests/init.pp`.
+> 
+> **Specifically**: Autoloading a function from a module does not trigger loading of the module's 
+> `manifests/init.pp` (nor is such initialization required to call a function from a module).
+> If an author of a module provides functions that require that the module's `manifests/init.pp`
+> is loaded, the function should include the module's class, or require that the caller first
+> includes the module's class).
+>
+> **Specifically**: a file `init.pp` under the `functions` directory of a module or the environment
+> does not have any special rules associated with it.
+> If that file exists it is supposed to contain a function named `<module>::init`.
+> Contrast this with `manifests/init.pp` which represents the module it is in. There is no such 
+> concept for functions. 
+
+### Defining functions in Manifests
+
+It is possible to define a Puppet Function in any manifest. Such functions will come into existence when the manifest in question is loaded for some reason other than calling the function (e.g. from 'manifests/site.pp' or when including a class).
+
+The following restrictions/rules/conventions apply on naming non-auto-loaded functions:
+
+* A function defined in a module **must** be qualified with the module's namespace
+* A function defined in an environment's main manifest **should** begin with the special namespace 'environment'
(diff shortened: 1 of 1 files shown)


language/puppet-functions.md
```diff
@@ -51,29 +51,60 @@ possible to call a puppet function with a lambda, and possibly also type a param
...
 Function Definition - Scope & Autoloading
 ---
-Puppet Functions are autoloaded from modules. They are located under <module-root>/functions.
-The filename must match the simple (non namespaced name part) of the function. Thus, a function 'min' in module 'testmodule' is placed like this:
+
+#### Autoloading
+
+Puppet Functions are autoloaded from modules and the environment when located under <module-root>/functions or <environment-root>/functions respectively.
+
+The filename must match the simple (non namespaced name part) of the function. Thus, a function 'min' in module 'testmodule' is located like this:
```
nfagerlund: > ...a function "testmodule::min" in module "testmodule" is located like...

Let's please never truncate namespaces when explaining this stuff. I know it probably seems perfectly clear to you, but it confuses readers.

language/puppet-functions.md
```diff
@@ -51,29 +51,60 @@ possible to call a puppet function with a lambda, and possibly also type a param
 
 Function Definition - Scope & Autoloading
 ---
-Puppet Functions are autoloaded from modules. They are located under <module-root>/functions.
-The filename must match the simple (non namespaced name part) of the function. Thus, a function 'min' in module 'testmodule' is placed like this:
+
+#### Autoloading
+
+Puppet Functions are autoloaded from modules and the environment when located under <module-root>/functions or <environment-root>/functions respectively.
```
nfagerlund: > Namespaced Puppet functions are autoloaded...

Better to specify like that right up front, since that is the only subset of functions affected by autoloading.

language/puppet-functions.md
```diff
@@ -51,29 +51,60 @@ possible to call a puppet function with a lambda, and possibly also type a param
...
 
 Rules:
 
-* autoloaded functions in modules must have qualified names
-* functions that are defined in modules, but that are not autoloaded (loaded as part of something   
-  else) should also be defined in the module's namespace (it should be an error if an attempt is 
-  made to name it differently).
-* functions defined in the environment may have any name (global or namespaced)
-* A function loaded in the environment shadows functions from modules, but not system functions
+* Files containing an auto loaded function may only contain a single function.
```
nfagerlund: Or else what?

language/puppet-functions.md
```diff
@@ -51,29 +51,60 @@ possible to call a puppet function with a lambda, and possibly also type a param
...
+* Auto loading a function from a module does not trigger loading of the module's `init.pp` (nor is such initialization required to call a function from a module). If an author of a module provides functions that require that the module's init.pp is loaded, the function should include the module's class, or require that the caller first includes the module's class).
```
nfagerlund: Do you mean that it never triggers loading of ANY manifests from the module's `manifests` dir?

Remember, now that this feature was added to the language, there are TWO possible directories full of manifests in a module: `manifests` and `functions`. We've lost the freedom to refer to manifest names without specifying what directory they are in.

language/puppet-functions.md
```diff
@@ -51,29 +51,60 @@ possible to call a puppet function with a lambda, and possibly also type a param
...
+
+##### Defining functions in Manifests
```
nfagerlund: Why is this a level 5 header? Do you consider it a subsection of the previous l4 header ("autoloading")? It doesn't read like a subsection of that.

language/puppet-functions.md
```diff
@@ -51,29 +51,60 @@ possible to call a puppet function with a lambda, and possibly also type a param
...
+
+It is possible to define a Puppet Functions in any manifest. Such functions will come into existence when the manifest in question is loaded for some reason other than calling the function (e.g. from 'site.pp' or when including a class).
+
+The following restrictions/rules/conventions apply on naming non-auto-loaded functions:
+
+* A function defined in a module **must** be qualified with the module's namespace
+* A function in an environment **should** begin with the special namespace 'environment'
+  * except when patching a function is required - then the name may shadow other functions defined in the same environment, or the modules in this environment, provided they have not already been auto loaded. Functions provided by the puppet runtime cannot be shadowed.
```
nfagerlund: How do you make sure the function has not been autoloaded? As a user, I have no idea how to do this safely and predictably.

language/puppet-functions.md
```diff
@@ -51,29 +51,60 @@ possible to call a puppet function with a lambda, and possibly also type a param
...
+* Files containing an auto loaded function may only contain a single function.
+* Auto loading a function from a module does not trigger loading of the module's `init.pp` (nor is such initialization required to call a function from a module). If an author of a module provides functions that require that the module's init.pp is loaded, the function should include the module's class, or require that the caller first includes the module's class).
+
+##### Defining functions in Manifests
+
+It is possible to define a Puppet Functions in any manifest. Such functions will come into existence when the manifest in question is loaded for some reason other than calling the function (e.g. from 'site.pp' or when including a class).
+
+The following restrictions/rules/conventions apply on naming non-auto-loaded functions:
+
+* A function defined in a module **must** be qualified with the module's namespace
```
nfagerlund: defined in a module's `manifests` dir?

language/puppet-functions.md
```diff
@@ -51,29 +51,60 @@ possible to call a puppet function with a lambda, and possibly also type a param
...
+* A function in an environment **should** begin with the special namespace 'environment'
```
nfagerlund: defined in an environment's main manifest?

nfagerlund:
This PR does not mention the magical `environment::` namespace, which to me was the single most confusing and surprising thing about this behavior. 

It also does not mention that the `functions/init.pp` file is never loaded, or explain why that's the case. (Or, if it is loadable, what circumstances cause it to be loaded.)

language/puppet-functions.md
hlindberg: ok, clearer.

language/puppet-functions.md
hlindberg: ok, clearer.

language/puppet-functions.md
hlindberg: or else error

language/puppet-functions.md
hlindberg: ok, it should be specific about 'manifests/init.pp' as opposed to a function 'function/init.pp' - will review all of what I wrote with that in mind.

language/puppet-functions.md
hlindberg: typo

language/puppet-functions.md
hlindberg: yes, but actually always true irrespective of where it is defined

language/puppet-functions.md
hlindberg: Because code that is part of initial import is loaded first; essentially all code "in the environment" (site.pp) is loaded first. If you follow the rules (that are currently not enforced) you will not run into trouble. I guess that what happens if you break the rules is not that interesting as you are not supposed to break them.

hlindberg:
I thought it did mention 'environment::' name space. Although in a terse way as this is a specification after all. Will re-read.

I don't understand the question about "function/init.pp" file never being loaded. Naturally if there is a function named mymodule::init(), and you call it, it will be loaded (from "mymodule/functions/init.pp". If you call another function mymodule::foo(), it will load the mymodule/functions/foo.pp file, but not the mymodule/functions/init.pp file. Is this perhaps a confusion over if just saying "init.pp" is a reference to "manifests/init.pp" or "functions/init.pp" (which is an ambiguity I had not considered). If so, then just clarifying that distinction should clear that up? Or, did I misunderstand the question?

Calling a function will load it if it isn't already loaded (i.e auto loaded if not already defined). Loading has no other side effect (nothing else is loaded; neither from manifests nor from functions).

nfagerlund:
Ah. You mention the `environment::` namespace under what "should" (but not "must") happen when loading functions from the _manifests_ dir, but you do not mention that it's mandatory in order to AUTOLOAD anything from an environment's `functions` dir. 

We need that info in the spec.

nfagerlund:
> Naturally if there is a function named mymodule::init(), and you call it, it will be loaded (from "mymodule/functions/init.pp".

This is a radical departure from the behavior of the normal (class / defined type) autoloader. In that autoloader, it is illegal to have a class named `<MODULE NAME>::init`, and `manifests/init.pp` should contain a class named `<MODULE NAME>`; however, if a class named `<MODULE NAME>::<SOMETHING>` is not found in `manifests/<SOMETHING>.pp`, init.pp will be loaded in an attempt to find that class. In (what I guess I'll call "the function autoloader, distinct from the normal autoloader"), init.pp—

* is not special
* will not be loaded when attempting to find a function named `<MODULE NAME>` or `environment`
* will not be loaded as a fallback when attempting to find a function named `<MODULE NAME>::<SOMETHING>` or `environment::<SOMETHING>`

This info is missing from the spec.

hlindberg:
Updated PR.

nfagerlund:
Excellent, this resolves all my problems. :+1:

igalic:
a few things are still kinda missing imo:
* is there a possibility to declare the return type?
* is the return value (like in ruby), simply the last value in the block?
* a full example, or two…

hlindberg:
Thanks igor; good points.
No return type declaration possible atm. And yes, last evaluated expression in any block/sequence is the result.

I may forget to update this in the spec unless there is a PUP task type ticket for it ;-)

- henrik

igalic:
@hlindberg [PUP-5602](https://tickets.puppetlabs.com/browse/PUP-5602)

PR #60, opened 2016-01-07
(PUP-5658) Disallow numeric types with ranges where to < from

thallgren:
Removes mention of descending ranges for Integer type from the
type specification.


hlindberg:
Removing as impl has deviated and we did not assert this detail. Also, PUP 5658 has a better more general solution.

PR #62, opened 2016-01-14
(PUP-5662)Specify that hash has keys in insertion order

hlindberg:


peterhuene:
:+1:

peterhuene:
Looks like there's this in `expressions.md`

> When matching against a Hash the keys are searched in an undefined order.

Although not sure why relevant, it would now be insertion order, correct?

hlindberg:
good catch, will look at that as well

hlindberg:
I updated the text. And since you gave thumbsup earlier I merged this.

The only possible relevance for the order (since `in` produces a boolean) is if you have an advanced reasoning about performance - it would take longer to answer if the last key is there than the first since the operation is sequential .  What you would do with this piece of information is however unclear :-)

peterhuene:
Sorry, didn't see it was updated.  Posthumous :+1:

PR #63, opened 2016-01-21
[WIP] (maint) Update windows file paths

melissa:


file_paths.md
```diff
@@ -208,111 +276,86 @@ create a `puppet` user or group.
...
-                ssleay32.dll *                # openssl dll
-                libeay32.dll *                # openssl dll
-                lib*.dll *                    # yaml, iconv, gdbi & ffi libraries
-            include *
-            lib *
-            share *
-
-        tools *
-            bin *
-              elevate.exe *                   # Used to elevate interactive commands
```
melissa: I'm not sure where this is being pulled in, but I can't find it.

file_paths.md
```diff
@@ -208,111 +276,86 @@ create a `puppet` user or group.
...
+                    puppet/*
+                    puppet_x.rb
+                    semver.rb
         modules *
-            pxp-module-puppet *
-            pxp-module-puppet.bat *
-
-    C:\Program Files\Puppet Labs\Puppet\service *
-        daemon.rb *                           # windows service daemon
-        nssm.exe *                            # NSSM used to run pxp-agent
```
melissa: This will also need to be installed, so we'll need to add an nssm.rb file to puppet-agent

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        routes.yaml                           # :route_file
+        ssl                                   # :ssldir
+
+    C:\ProgramData\PuppetLabs\pxp-agent
+        modules *                             # stores configuration files for pxp-agent modules
+            pxp-module-puppet.conf *          # configuration file of the pxp module puppet (optionally, to override puppet.bat location)
+        pxp-agent.conf                        # pxp-agent configuration file
+
+    C:\Program Files\Puppet Labs\bin *
+        environment.bat *                     # setup LOAD_PATH
```
MikaelSmith: Would expect to remove environment.bat.

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        puppet.conf *                         # :config
+        routes.yaml                           # :route_file
+        ssl                                   # :ssldir
+
+    C:\ProgramData\PuppetLabs\pxp-agent
+        modules *                             # stores configuration files for pxp-agent modules
+            pxp-module-puppet.conf *          # configuration file of the pxp module puppet (optionally, to override puppet.bat location)
+        pxp-agent.conf                        # pxp-agent configuration file
+
+    C:\Program Files\Puppet Labs\bin *
```
melissa: I'm trying to figure out the best way to handle these. We can either copy or move the ${project}.bat files from `C:\Program Files\Puppet Labs\puppet\bin`, or figure out linking? Or tell install.rb to install them here. Any non-project-specific files (like environment.bat) should live in puppet-agent while the rest should live in their corresponding projects

cc/ @MikaelSmith

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+            ruby.exe *
+            rubyw.exe *
+            service_manager.rb ?
+            ssleay32.dll *                    # openssl dll
+            stompcat *
+            stompcat.bat *
+            testrb.bat *
+            unregister_service.bat ?
+            x64-msvcrt-ruby210.dll *          # ruby dll
+            *.dll *                           # dependency libraries (zlib, gcc, mingw, boost, etc.)
```
melissa: For ruby before, we were installing a bunch of supporting libraries (yaml, iconv, gdbi & ffi, etc), but we're not anymore. Should we be copying those over in the runtime project (like we are with zlib and gcc libraries)?

MikaelSmith: What might be needed for Ruby DevKit?

file_paths.md
MikaelSmith: Symlinking should work on Windows, so I'd give that a try.

file_paths.md
melissa: There might be some other things we set there that we need to carry forward. I'm including it in this list just for old times sake, really

file_paths.md
melissa: :+1: awesome, I'll look into that then

file_paths.md
```diff
@@ -152,7 +152,7 @@ The files annotated by an '*' indicate that they are created by package installa
 
 On recent versions of Windows, e.g. 2008 & 2012, the installation path
 defaults to `C:\Program Files\Puppet Labs` and the common app data
-directory defaults to `C:\ProgramData`. On 2003, common app data is
+directory defaults to `C:\ProgramData\PuppetLabs`. On 2003, common app data is
```
joshcooper: Nit, "common app data" is a special windows specific folder and is `C:\ProgramData`, and we add our stuff within `PuppetLabs`.

file_paths.md
joshcooper: For strange reasons, only administrators running elevated can create symlinks on windows, so that could be problematic in non-admin installs. Fortunately we don't need to worry about 2003 (as that doesn't support symlinks).

file_paths.md
joshcooper: For "reasons" it's in https://github.com/puppetlabs/puppet-win32-ruby/tree/2.1.x-x64/tools/bin. Note the `elevate.exe.config` needs to be installed in the same directory. Ideally we can put on buildsources? The executable is used when selecting some of the shortcuts, e.g. https://docs.puppetlabs.com/puppet/latest/reference/services_commands_windows.html#run-puppet-agent.

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+            libfacter.so *
+            mco *                             # ruby bin wrapper
+            mco.bat * (should this be moved?)
+            mcollectived *                    # ruby bin wrapper
+            minitar *
+            minitar.bat *
+            openssl.exe *
+            puppet *                          # ruby bin wrapper
+            puppet.bat * (should this be moved?)
+            puppetres.dll ?
```
joshcooper: This is puppet's [message file](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363669\(v=vs.85\).aspx) resource dll. It's used to format puppet log messages in the Windows event viewer. I compiled it along time ago...

Currently, we always format the messages as en_US, but in theory, if puppet were to output a unique message ID for each message, then event viewer could localize the message text by using the appropriate message file resource dlls.

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+            pxp-agent.exe *
+            rake.bat *
+            rdoc.bat *
+            register_service.bat ?
```
joshcooper: This comes from mco and is used if you wanted to manually register the mco service. I don't think anything in puppet uses it (since we register the service in the WiX definition).

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+            ri.bat *
+            ruby.exe *
+            rubyw.exe *
+            service_manager.rb ?
```
joshcooper: Same mco comment. I think it's only used when trying to run mco from source on Windows, and can probably be omitted.

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+            ssleay32.dll *                    # openssl dll
+            stompcat *
+            stompcat.bat *
+            testrb.bat *
+            unregister_service.bat ?
```
joshcooper: Same mco comment.

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
-    -------------------------------------------------------------------------------------------------------
-        cache                                 # :vardir                      C:\ProgramData\PuppetLabs\puppet\var
+    C:\Program Files\Puppet Labs\puppet *     # ruby-puppet root
+        bin *                                 # executables and dlls
+            c_rehash *
+            catstomp *
+            catstomp.bat *
+            curl.exe *
+            curl-config *
+            daemon.bat ?
```
joshcooper: Can be removed. In 3.x, we used to install the puppet service and have it execute [daemon.bat](https://github.com/puppetlabs/puppet_for_the_win/blob/3.7.5/wix/puppet.wxs#L162). We changed this at some point so the service control manager effectively calling [`ruby.exe daemon.rb`](https://github.com/puppetlabs/puppet_for_the_win/blob/stable/wix/puppet.wxs#L160-L173).

file_paths.md
joshcooper: Sadly that script does a bunch of things. It might be good to ticket moving that logic elsewhere so that we can remove it in the future?

* sets [env_windows_installdir](https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L8) fact. Could be a core fact instead, since many things rely on it, e.g. PE.
* builds up [PATH](https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L25). Hopefully all of that logic goes away in the layout?
* sets [RUBYLIB](https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L27-L31). I think that logic can go away because ruby code is in `vendor_ruby`. Also I don't know that we need to convert to forward slashes any more.
* might be another way to [enable rubygems support](https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L33-L35)?
* overrides [trusted ssl directories](https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L37-L39)

file_paths.md
melissa: Good to know, I'll change this back

file_paths.md
melissa: I actually just realized this is coming from mcollective (and how I'm installing it). https://github.com/puppetlabs/marionette-collective/tree/master/ext/windows see https://github.com/puppetlabs/marionette-collective/pull/363

file_paths.md
melissa: We should probably add a new component file for this, unless it can be categorized under ruby or something else. But yeah, we can totes drop it into buildsources and make sure `elevate.exe.config` is in the same directory

file_paths.md
melissa: They probably are needed for ruby devkit. But are we supporting the user to use that on their system? Should those be reserved for puppet-client-tools or something that's more dev specific?

file_paths.md
joshcooper: Oh right! Puppet's `daemon.bat` was in `/service` but that was removed awhile ago. We can't switch the `wix` file over to calling `ruby.exe mcollectived.rb` because the `mcollectived` service needs to load all of the mco code (so needs to call out to `environment.bat` to setup `RUBYLIB`, etc).

file_paths.md
MikaelSmith: The notes I found on it were http://stackoverflow.com/questions/14141436/is-there-any-way-to-create-symbolic-link-in-wix-installer and http://stackoverflow.com/questions/3611501/wix-make-symbolic-link-with-uac-enabled, in case they add any useful context.

file_paths.md
melissa: So, if we're installing the mco libs in a place where ruby can find them, we shouldn't need to call environment.bat, right? So, we'd then be able to call `ruby.exe mcollectived.rb`?

file_paths.md
melissa: Hmm, where is this sourced? It's getting installed just fine, so it's probably fine? Is this a fine place for it to be installed?

file_paths.md
melissa: So, in the list of files from https://github.com/puppetlabs/marionette-collective/tree/master/ext/windows, we only need mco.bat? Everything else shouldn't be necessary except for environment.bat, but we'll have a project-level version of that file that will work with everything (a la puppet_for_the_win). Does that sounds right?

file_paths.md
joshcooper: Currently comes from puppet: https://github.com/puppetlabs/puppet/tree/master/ext/windows/eventlog. It would be great to move that elsewhere and make it available on buildsources.

file_paths.md
joshcooper: I think so... We'll need to make sure the wix file runs the mco service using `rubyw.exe` like the `daemon.bat` script does.

file_paths.md
```diff
@@ -163,39 +163,107 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+            environment.bat ?
+            erb.bat *
+            extlookup2hiera *
+            extlookup2hiera.bat *
+            facter.exe *
```
melissa: Will we also need a `facter.bat`?

file_paths.md
```diff
@@ -208,111 +276,86 @@ create a `puppet` user or group.
...
+                openssl.pc
+                ruby-2.1.pc
+            ruby
+                2.1.0
+                gems
+                    2.1.0
+                        specifications
+                            facter.gemspec
+                            hiera.gemspec
+                            puppet.gemspec
```
melissa: Should we also have an `mcollective.gemspec`?

file_paths.md
```diff
@@ -208,111 +276,86 @@ create a `puppet` user or group.
...
+            aclocal
+            doc
+            man
+            vim
+            zsh
+        ssl *
+            cert.pem
+            certs
+            openssl.cnf
+        var *
```
melissa: I wasn't sure where to drop the `log` and `run` directories. I just dropped them here...

file_paths.md
```diff
@@ -163,156 +163,166 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+    C:\Program Files\Puppet Labs\facter *
         facts.d *                             # external facts directory (not pluginsync'ed)
 
-    C:\ProgramData\PuppetLabs\mcollective
+    C:\Program Files\Puppet Labs\mcollective *
         plugins *                             # user installed plugins
-        etc *                                                                same
-            client.cfg
-            facts.yaml
-            server.cfg
```
melissa: Note, these were moved from `C:\ProgramData\PuppetLabs\mcollective\etc` to `C:\ProgramData\PuppetLabs\mcollective` and the plugins directory was moved from `C:\ProgramData\PuppetLabs\mcollective\plugins` to `C:\Program Files\Puppet Labs\mcollective\plugins`. If this is wrong, we can change it

file_paths.md
```diff
@@ -163,156 +163,166 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        hiera.yaml                            # :hiera_config
+        modules                               # user modulepath
+
+    C:\ProgramData\PuppetLabs\mcollective *
+        client.cfg *
+        facts.yaml *
+        server.cfg *
 
-    C:\ProgramData\PuppetLabs\facter *                                       same
+    C:\ProgramData\PuppetLabs\puppet          # :confdir
```
melissa: This was moved from `C:\ProgramData\PuppetLabs\puppet\etc`

file_paths.md
```diff
@@ -163,156 +163,166 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        binder_config.yaml                    # :binder_config
+        csr_attributes.yaml                   # :csr_attributes
+        custom_trusted_oid_mapping.yaml       # :trusted_oid_mapping_file
+        device.conf                           # :deviceconfig
+        fileserver.conf                       # :fileserverconfig
+        puppet.conf *                         # :config
+        routes.yaml                           # :route_file
+        ssl                                   # :ssldir
+
+    C:\ProgramData\PuppetLabs\pxp-agent
```
melissa: This was moved from `C:\ProgramData\PuppetLabs\pxp-agent\etc`

file_paths.md
```diff
@@ -163,156 +163,166 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+          ruby.exe *
+          rubyw.exe *
+          ssleay32.dll *                      # openssl dll
+          stompcat *
+          stompcat.bat *
+          testrb.bat *
+          x64-msvcrt-ruby210.dll *            # ruby dll
+          *.dll *                             # dependency libraries (zlib, gcc, mingw, boost, etc.)
+
+        cache *                               # :vardir                      C:\ProgramData\PuppetLabs\puppet\var
```
melissa: This was moved from `C:\ProgramData\PuppetLabs\puppet\cache` to `C:\Program Files\Puppet Labs\puppet\cache`

file_paths.md
```diff
@@ -163,156 +163,166 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
-        var *
-            log *
-                pxp-agent.log                 # enabled by default
-            spool *                           # directory containing results of pxp-agent modules
-            run *
-
-    C:\Program Files\Puppet Labs\Puppet
-        VERSION                               # puppet-agent package version
-
-    C:\Program Files\Puppet Labs\Puppet\bin *
```
melissa: This was moved to `C:\Program Files\Puppet Labs\bin`

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        hiera.yaml                            # :hiera_config
+        modules                               # user modulepath
 
-    C:\ProgramData\PuppetLabs\facter *                                       same
+    C:\ProgramData\PuppetLabs\mcollective *
+        client.cfg *
+        facts.yaml *
+        server.cfg *
+
+    C:\ProgramData\PuppetLabs\puppet          # :confdir
```
melissa: This was moved from `C:\ProgramData\PuppetLabs\puppet\etc`

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
             manifests *
             modules *
-        hiera.yaml *                          # :hiera_config
-        hieradata *                           # n/a
-        modules *                             # user modulepath
+        hiera.yaml                            # :hiera_config
+        modules                               # user modulepath
 
-    C:\ProgramData\PuppetLabs\facter *                                       same
+    C:\ProgramData\PuppetLabs\mcollective *
```
melissa: This was moved from `C:\ProgramData\PuppetLabs\mcollective\etc`

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        environment.bat@ *                    -> C:\Program Files\Puppet labs\puppet\bin\environment.bat
+        puppet.bat@ *                         -> C:\Program Files\Puppet labs\puppet\bin\puppet.bat
+        hiera.bat@ *                          -> C:\Program Files\Puppet labs\puppet\bin\hiera.bat
+        mco.bat@ *                            -> C:\Program Files\Puppet labs\puppet\bin\mco.bat
+        facter.bat@ *                         -> C:\Program Files\Puppet labs\puppet\bin\facter.bat
+        puppet_shell.bat@ *                   -> C:\Program Files\Puppet labs\puppet\bin\puppet_shell.bat
+        run_facter_interactive.bat@ *         -> C:\Program Files\Puppet labs\puppet\bin\run_facter_interactive.bat
+        run_puppet_interactive.bat@ *         -> C:\Program Files\Puppet labs\puppet\bin\run_puppet_interactive.bat
+
+    C:\Program Files\Puppet Labs\facter *
```
melissa: This was moved from `C:\ProgramData\PuppetLabs\facter`

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+            hieradata *
             manifests *
             modules *
-        hiera.yaml *                          # :hiera_config
-        hieradata *                           # n/a
-        modules *                             # user modulepath
+        hiera.yaml                            # :hiera_config
+        modules                               # user modulepath
 
-    C:\ProgramData\PuppetLabs\facter *                                       same
```
melissa: This was moved to `C:\Program Files\Puppet Labs\facter`

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        puppet.conf *                         # :config
+        routes.yaml                           # :route_file
+        ssl                                   # :ssldir
+
+    C:\ProgramData\PuppetLabs\pxp-agent
+        modules *                             # stores configuration files for pxp-agent modules
+            pxp-module-puppet.conf *          # configuration file of the pxp module puppet (optionally, to override puppet.bat location)
+        pxp-agent.conf                        # pxp-agent configuration file
+
+    C:\ProgramData\PuppetLabs\var
```
melissa: Consolidates all log/pid files into one location

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        log *
+          mcollective.log
+          puppet.log
+          pxp-agent.log
+        run *
+          agent.pid
+          mcollective.pid
+          pxp-agent.pid
+
+    C:\Program Files\Puppet Labs\bin *        # symlink targets of puppet related binaries
```
melissa: Moved from ` C:\Program Files\Puppet Labs\Puppet\bin`

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+        run_facter_interactive.bat@ *         -> C:\Program Files\Puppet labs\puppet\bin\run_facter_interactive.bat
+        run_puppet_interactive.bat@ *         -> C:\Program Files\Puppet labs\puppet\bin\run_puppet_interactive.bat
+
+    C:\Program Files\Puppet Labs\facter *
         facts.d *                             # external facts directory (not pluginsync'ed)
 
-    C:\ProgramData\PuppetLabs\mcollective
+    C:\Program Files\Puppet Labs\mcollective *
         plugins *                             # user installed plugins
-        etc *                                                                same
```
melissa: Moved to `C:\ProgramData\PuppetLabs\mcollective`

melissa:
PR for MCO work is at https://github.com/puppetlabs/marionette-collective/pull/363 and awaiting review

file_paths.md
melissa: https://tickets.puppetlabs.com/browse/RE-6123 for the linking stuff

ferventcoder:
This is going to require lots of testing once it is implemented. 

For instance, how does one take the existing conf file and move it during upgrade?

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
     C:\ProgramData\PuppetLabs\client-tools    # default client tool settings
         puppet-access.conf *
         puppet-orchestrator.conf *
 
-
     C:\ProgramData\PuppetLabs\code *          # :codedir                     contents moved from C:\ProgramData\PuppetLabs\puppet\etc (:confdir)
         environments *                        # :environmentpath
           production *
             environment.conf *
+            hieradata *
```
ferventcoder: Wasn't this already available here?

ferventcoder: Keep in mind that last time we moved around the hieradata and changed the hiera.yaml, it was a major version bump for hiera. This alone may cause the Puppet Agent to need a major version bump.

file_paths.md
ferventcoder: So we are dropping the etc folder then?

ferventcoder: The agent will need to detect and move existing files.

file_paths.md
ferventcoder: The agent will need to detect and move existing files.

file_paths.md
ferventcoder: Same as above.

file_paths.md
ferventcoder: :+1:

file_paths.md
ferventcoder: :+1:

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
-        LICENSE.rtf *                         # license
-        puppetlabs.ico *                      # icon for start menu shortcut
-        puppetres.dll *                       # event log message resource dll
-        versions.txt *                        # versions of components
+    C:\Program Files\Puppet Labs\puppet *     # ruby-puppet root
+        bin *                                 # executables and dlls
+          c_rehash *
+          catstomp *
+          catstomp.bat *
+          curl.exe *
```
ferventcoder: How did this sneak in here?

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
+          curl-config *
+          environment.bat *                   # setup LOAD_PATH, globally used by all our .bat files, to be symlinked to C:\Program Files\Puppet Labs\bin
+          erb.bat *
+          extlookup2hiera *
+          extlookup2hiera.bat *
+          facter.exe *                        # Facter executable (we need to add facter.bat also I think)
+          gem.bat *
+          hiera *                             # ruby bin wrapper
+          hiera.bat *                         # To be symlinked to C:\Program Files\Puppet Labs\bin
```
ferventcoder: :+1:

file_paths.md
ferventcoder: You'll need a plan to upgrade this as well.

file_paths.md
ferventcoder: Thought we already had one?

file_paths.md
ferventcoder: I don't see `C:\Program Files\Puppet Labs\Puppet\bin` anymore, but in each of the files mco.bat, puppet.bat, etc you mention they will symlink to `C:\Program Files\Puppet Labs\Puppet\bin`, does the directory need to continue to exist?

file_paths.md
ferventcoder: Typically you want to keep user data out of Program Files - that is what ProgramData is for. Let's be sure facts.d isn't something we would consider user data.

Basically the argument is that if files/directories are expected to be added/changed after install - they go to ProgramData. cc @jpogran @joshcooper

file_paths.md
melissa: purposefully... https://github.com/puppetlabs/puppet-agent/blob/feature/master/windows-vanagon-development/configs/components/curl.rb the curl executable is also available in *nix puppet-agent packages

file_paths.md
melissa: We have nssm.json, but vanagon needs to know how to process the nssm archive. Previously puppet_for_the_win was taking care of that.

file_paths.md
melissa: Coolio. This isn't arch-specific or anything, right? Also, I'm fine getting it up on buildsources, but I'd prefer to have it versioned. I'll put it up now in the meantime though

file_paths.md
melissa: If we're doing a major upgrade everytime, is this still the case? Or is this because the config files are left behind?

ferventcoder: ProgramData files are not touched.

ferventcoder: They are created in ProgramData if they do not exist, but otherwise they are completely left alone.

MikaelSmith:
In general config files shouldn't be moving.

file_paths.md
MikaelSmith: I agree, this looks like a mistake, unless we're correcting a previous mistake in the specification.

file_paths.md
MikaelSmith: Agree with @ferventcoder, config file locations shouldn't change. `facts.d` should still be in `C:\ProgramData\PuppetLabs\facter\facts.d`.

file_paths.md
MikaelSmith: We shouldn't be moving the `plugins` location.

file_paths.md
```diff
@@ -163,156 +163,167 @@ and `pxp-agent`, all running as `LocalSystem` by default. It will not
...
-            client.cfg
-            facts.yaml
-            server.cfg
-        var *                                                                same
-            log *
-                mcollective.log
-
-    C:\ProgramData\PuppetLabs\puppet
-    -------------------------------------------------------------------------------------------------------
-        cache                                 # :vardir                      C:\ProgramData\PuppetLabs\puppet\var
```
MikaelSmith: This is all config or runtime data, location should not change.

file_paths.md
MikaelSmith: Hmm, was hieradata ever used on Windows?

file_paths.md
joshcooper: We shouldn't be moving this to "Program Files". The cache is data that belongs in `ProgramData`, and not with binaries that live in `Program Files`.

file_paths.md
joshcooper: Correct, not arch specific, but where does it live? It looks like this PR deletes it.

file_paths.md
joshcooper: `hieradata` would have only been used if you were running `puppet apply` on Windows.

ferventcoder: Some folks do, which could make this a breaking change if not handled during an upgrade.

file_paths.md
melissa: Cool, I'll change it back

file_paths.md
melissa: I can change it back

file_paths.md
melissa: Ooh, cool, I'll change this back then

file_paths.md
melissa: sure thing, I'll change it back

file_paths.md
melissa: @ferventcoder are you asking if `C:\Program Files\Puppet Labs\bin` or `C:\Program Files\Puppet Labs\puppet\bin` need to exist? `C:\Program Files\Puppet Labs\puppet\bin` will contain ruby, third party executables and dlls, etc. Whereas `C:\Program Files\Puppet Labs\bin` will contain the batch files for our projects. That is, if I can figure out symlinking...

file_paths.md
melissa: Sure, I'll change them back

file_paths.md
melissa: Cool, I'll change it back

melissa:
I'll try to get a PR up next week sometime against the windows feature branch in puppet-agent and update this to reflect that

file_paths.md
melissa: Does nssm.exe provide daemon.rb? Where does that come from?

file_paths.md
melissa: `C:\Program Files\Puppet Labs\puppet\elevate.exe`. I didn't add it in until just now since I didn't know where it should be

melissa:
Updated with https://github.com/puppetlabs/puppet-agent/pull/524

file_paths.md
```diff
@@ -220,99 +215,132 @@ create a `puppet` user or group.
...
-        lib *
-            hiera.rb *
 
-    C:\Program Files\Puppet Labs\Puppet\mcollective *
-        bin *
-            mcollectived *                    # ruby bin wrapper
-            mco *
-        lib *
-            mcollective.rb *
+    C:\ProgramData\PuppetLabs\var
```
melissa: @ferventcoder I think you gave this move a thumbs up, but since it's in ProgramData, is it something we can do? Or do these pidfiles/logfiles need to be where they originally were?

file_paths.md
ferventcoder: We could just drop the bat files there and punt on symlinking, as suggested earlier by @joshcooper. Not sure where this may be now, so this may be stale information.

ferventcoder:
@puppetlabs/windows

file_paths.md
ferventcoder: Probably should keep all existing ProgramData where they were for now.

file_paths.md
```diff
@@ -220,99 +215,132 @@ create a `puppet` user or group.
...
+            puppet *
+                puppet.log
+            pxp-agent *
+                pxp-agent.log
+        run *
+            agent.pid
+            mcollective.pid
+            pxp-agent.pid
+
+    C:\Program Files\Puppet Labs\bin *        # symlink targets of puppet related binaries
```
melissa: @jpogran will this work? It's outside `[INSTALLDIR]`...

melissa: @jpogran will this work? It's outside `[INSTALLDIR]`...

file_paths.md
melissa: Addressed in https://github.com/puppetlabs/puppet-agent/pull/531

file_paths.md
```diff
@@ -237,84 +234,122 @@ create a `puppet` user or group.
...
-        facter.bat *                          # bat file wrappers
-        hiera.bat *
-        mco.bat *
-        puppet.bat *
-        puppet_shell.bat *                    # targets for shortcuts
-        run_facter_interactive.bat *
-        run_puppet_interactive.bat *
-
-    C:\Program Files\Puppet Labs\Puppet\facter *
+    C:\Program Files\Puppet Labs\bin *        # symlink targets of puppet related binaries
```
melissa: @jpogran will this work installing things outside of `[INSTALLDIR]`?

file_paths.md
melissa: nvm, found it, installing into bindir

file_paths.md
MikaelSmith: This can only be the default location... the MSI could install `C:\Program Files\Puppet Labs` to anywhere (like localized `C:\Program Files`).

file_paths.md
```diff
@@ -237,84 +234,122 @@ create a `puppet` user or group.
...
+        environment.bat@ *                    -> C:\Program Files\Puppet labs\Puppet\bin\environment.bat
```
MikaelSmith: These will have to be relative (either symlinks or batch files invoking the targets).

file_paths.md
```diff
@@ -237,84 +234,122 @@ create a `puppet` user or group.
...
+            c_rehash *
+            catstomp *
+            catstomp.bat *
+            curl.exe *
+            curl-config *
+            daemon.bat *
+            daemon.rb *                         # windows service daemon
+            elevate.exe *                       # Used to elevate interactive commands
+            elevate.exe.config *
+            environment.bat *                   # setup LOAD_PATH, globally used by all our .bat files, to be symlinked to C:\Program Files\Puppet Labs\bin
```
MikaelSmith: Not sure why it needs to be symlinked to `Puppet Labs\bin`.

file_paths.md
```diff
@@ -237,84 +234,122 @@ create a `puppet` user or group.
...
+            facter *                            # facter headers
+            leatherman *
+            openssl *
+            ruby-2.1.0 *
 
-    C:\Program Files\Puppet Labs\Puppet\misc *
-        LICENSE.rtf *                         # license
-        puppetlabs.ico *                      # icon for start menu shortcut
-        puppetres.dll *                       # event log message resource dll
-        versions.txt *                        # versions of components
```
shrug: is this file going away?

melissa: Nope, I just haven't figured out where to drop it yet. We're going to use `bill-of-materials` to replace it.

file_paths.md
melissa: So it could be one up from `INSTALLDIR`? In this example, we assume `[INSTALLDIR] = C:\Program Files\Puppet Labs\Puppet`. Given that this is one directory level above `Puppet`, outside of `[INSTALLDIR]`, I'm wondering if this location needs to be rethought

file_paths.md
jcoconnor: Versions.txt needs to be included - should we continue with a misc directory for this purpose ?
Also - I think it is advisable to include the LICENSE.rtf file.

melissa: @jcoconnor PR accepted

file_paths.md
MikaelSmith: That... seems problematic. There's also `C:\Program Files\Puppet Labs\pxp-agent` below, which has the same problem.

file_paths.md
joshcooper: if puppetres.dll is moved, please make sure to update the registry key that registers it as an event viewer source.

file_paths.md
melissa: Yeah, that's why I brought it up. I'm thinking we could have something like:
```
C:\Program Files\Puppet Labs\Puppet\bin
C:\Program Files\Puppet Labs\Puppet\puppet
C:\Program Files\Puppet Labs\Puppet\puppet\bin
C:\Program Files\Puppet Labs\Puppet\pxp-agent
```
That way the batch files are in the same place they've always been without having to worry about conflict with things like `facter.exe`

MikaelSmith: That's not the worst.

P.S. The file type's the worst.

melissa: https://github.com/puppetlabs/puppet-agent/pull/542 to update the path to what I commented on above

file_paths.md
```diff
@@ -163,160 +163,195 @@ The package will create the following services `puppet`, `mcollective`,
 and `pxp-agent`, all running as `LocalSystem` by default. It will not
 create a `puppet` user or group.
 
-    Path                                      Setting                        3.x
-    C:\ProgramData                                                           n/a
+    Path                                          Setting                        3.x
+    C:\ProgramData\PuppetLabs *                                                  n/a
```
jcoconnor: @melissa Is there a reason why we are setting this to ```C:\ProgramData\PuppetLabs``` whereas the Program Files are being installed under: ```C:\Program Files\Puppet Labs``` ?

Just asking because @McdonaldSeanp and I are running through the installed directories at the moment to make sure everything is consistent.

melissa: I couldn't get vanagon to build with spaces in the path name. So, from what I last saw in the vanagon build, it was building under `PuppetLabs`.

MikaelSmith: This was always `C:\ProgramData\PuppetLabs`. It's used for storing settings, not the installed application.

melissa: Good point! It was `C:\Program Files\Puppet Labs` that I had to change to  `C:\Program Files\PuppetLabs`

joshcooper: It would be great to figure out how to use the original install path with spaces, otherwise there will be downstream docs changes, etc.

file_paths.md
jcoconnor: @melissa We still don't have a location for these 3 files - can I re-instate the ```misc``` directory for these unless you have a better suggestion ?
```
LICENSE.rtf
versions.txt
puppetlabs.ico
```

melissa: I have no qualms with that

jcoconnor: Super - I'll run with that - can you update the spec in this PR please ?

melissa: done

file_paths.md
```diff
@@ -163,160 +163,202 @@ The package will create the following services `puppet`, `mcollective`,
...
-        puppet_shell.bat *                    # targets for shortcuts
-        run_facter_interactive.bat *
-        run_puppet_interactive.bat *
-
-    C:\Program Files\Puppet Labs\Puppet\facter *
-        bin *                                 # executables and dlls
+    C:\Program Files\Puppet Labs\Puppet\misc *
+        LICENSE.rtf *                             # license
+        puppetlabs.ico *                          # icon for start menu shortcut
+        puppetres.dll *                           # event log message resource dll
```
jcoconnor: @melissa Just wondering if we should leave ```puppetres.dll``` in the ```bin``` directory (where it is built to now).
Otherwise I will need to look at the build and how this is built.
The ```.ico``` file can live either here or bin directory, but since the only other files here are "document type", i.e. license and version list, it would probably be better to put the ```.ico`` file in ```bin```

Tx

melissa: I don't have an opinion. I just copied/pasta'd this here. I know @joshcooper was saying if we move this, we need to update the registry entry though. You can always fork this branch and make any changes you think are needed. I can close this PR if you'd like to open one from your branch?

jcoconnor: yeah - I have a PR open to handle the .dll anyways.
Was curious as to the how this .dll is built, so have taken a build log to look at in the morning.
I also might take up your offer to fork and take over the PR if its ok with you.

jcoconnor:
Closing to continue this with PR https://github.com/puppetlabs/puppet-specifications/pull/70

PR #65, opened 2016-02-05
(PUP-5772) Specify the Iterable and Iterator types.

hlindberg:
This adds the Iterable and Iterator types to the specification. They are
introduced in Puppet 4.4.0.


igalic:
Do i understand this correctly, that developers can create their own `Iterator`s?

hlindberg:
@igalic Yes basically - we have the type `Iterator[T]` which makes it possible to return an iterator to puppet - see the implementation of the `step` function in puppet as an example.

PR #66, opened 2016-02-08
(PUP-5836) Add specificatio of splat iterator to array

hlindberg:
This also improves the text for the Unary Splat operator


hlindberg:
ping @thallgren - please review and merge

hlindberg:
the long lines depends on the tool being used to edit  (I use Mou) and there you cannot see the long lines. A problem with wrapping them is that you get a lot of changes in the diff when paragraphs
are re-wrapped. Not sure that we care in files that are not intended to be viewed raw.

It is easier to read the text in a review when paragraphs are broken up on lines though.

hlindberg:
@thallgren the typo is fixed now.

PR #72, opened 2016-04-19
Document current resource type API

joshcooper:
Documents the current resource type API in ruby with an emphasis on the
lack of environment isolation provided by the current implementation.


language/resource_types.md
```diff
@@ -0,0 +1,440 @@
+# Puppet Resource Types
+
+This document describes puppet's DSL for defining puppet resource types. Examples are taken from the `puppetlabs-aws` module and core puppet.
```
hlindberg: Small, thing, but think we should say "puppet's Ruby DSL for defining resource types" to avoid confusion with the "puppet DSL" a.k.a "the puppet language", and to avoid confusion with "types" (i.e. data types in general. Later in the text I am fine using just 'type' as a term.

language/resource_types.md
```diff
@@ -0,0 +1,440 @@
...
+    [ [ /^(.*?)\/*\Z/m, [ [ :path ] ] ] ]
+  end
+end
+```
+
+A type must define a `title_patterns` method when using composite namevars so that puppet knows how to decompose the `title` into its constituent namevars, e.g. see [java-ks](https://github.com/puppetlabs/puppetlabs-java_ks/blob/abac95473a505080f60b4d0118c82d3568063da0/lib/puppet/type/java_ks.rb#L150-L176), [websphere](https://github.com/puppetlabs/puppetlabs-websphere_application_server/blob/cd2dbc59c43030db36a81f9c1ed155d5fbe4e85c/lib/puppet/type/websphere_sdk.rb#L11-L28).
+
+### ensureable
+
+Creates an `ensure` property with acceptable values of `present` and `absent`, each of which invoke the provider's `create` and `destroy` methods, respectively:
```
hlindberg: Question: is it possible to have an `ensure` that does not include `present` and `absent` as values?

hlindberg:
Excellent write up!

language/resource_types.md
joshcooper: Good point will update.

language/resource_types.md
joshcooper: Like many things in puppet, it depends...

You could define your own `ensure` property with custom values, e.g.

```ruby
newproperty(:ensure) do
  newvalues(:red, :green)
end
```

Though I don't know if that would actually work, as `:ensure` gets [special treatment](https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/transaction/resource_harness.rb#L77) in various parts of puppet. For example, if a provider returns the current value of a property, `nil` and `:absent` are usually treated as the same.

If you call `[Puppet::Type.ensurable](https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/type.rb#L203)` without a block then you get the default `:present` and `:absent` values and behavior.

If you call it with a block, like the [package type does](https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/type/package.rb#L69-L106), then you can define whatever values you want. It just so happens the package type defines `:present`, `:absent`, and also `:purged`, `:held`, `:latest`, etc. But I don't know if that's true for other custom types.

joshcooper: I added a section about ensurable with a block.

ripienaar:
No specific comment other than to say this is excellent, we've not had docs about this stuff - this is a good document, hopefully we get similar for providers!

language/resource_types.md
hlindberg: Thanks Josh - I asked because I wondered if a new implementation needs to enforce that ensurable always is at least absent or present. The addition you made here is clear about what it does now. The rest is a discussion that we can have someplace else.

adrienthebo: We use custom ensure values for the service type as well: https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/type/service.rb#L84-L112

hlindberg: Are those in addition to the base values 'absent' and 'present' ? @adrienthebo

joshcooper: If a block is passed, then the values specified are the complete set, which may not include the default absent/present values:

```
irb(main):011:0> Puppet::Type.type(:service).attrclass(:ensure).value_collection.values
=> [:stopped, :running]
irb(main):012:0> Puppet::Type.type(:file).attrclass(:ensure).value_collection.values
=> [:absent, :file, :present, :directory, :link]
irb(main):013:0> Puppet::Type.type(:package).attrclass(:ensure).value_collection.values
=> [:present, :absent, :purged, :held, :latest]
```

adrienthebo: For what it's worth, the fact that the package ensure value can `ensure => held` is deeply problematic because you can't hold a package at a specific version.

Oh, and then there's that - the package type also accepts a version as an ensure value, which amounts to it accepting an arbitrary string.

This is bad and terrible and makes me quite unhappy. I would love if we gradually deprecated these things and moved to ensure values being `present` and `absent` and everything else being shoved into a property.

hlindberg:
I am fine with merging this. Any reason to continue having it up for review?

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+  newproperty(:enable, :required_features => :enableable) do
+  end
+end
+```
+
+A **provider** indicates it supports the feature using `has_feature`:
+
+```ruby
+Puppet::Type.type(:service).provide :launchd, :parent => :base do
+  has_feature :enableable
```
nanliu: Can a provider require features based on resource attributes? For example require 
```
archive { 'some_file'
  source => 's3:// ...',
}
```
Depending on the source, the provider should require has_feature s3, has_feature puppet://, has_feature http://, etc.

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+Puppet parameter and properties can be multi-valued. For example, the `ec2_instance` can be given a list of `security_groups`, and each value in the list will be yielded to the `validate` method:
+
+```ruby
+Puppet::Type.newtype(:ec2_instance) do
+  newproperty(:security_groups, ...) do
+    validate do |value|
+      fail 'security_groups should be ...' if value !~ /something/
+    end
+  end
+end
```
nanliu: There's still global validate that's not tied to specific property.

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+        fail 'tags should be a Hash' unless value.is_a?(Hash)
+      end
+    end
+  end
+end
+```
+
+###  boolean
+
+Meta-programs a predicate method for the parameter/property. For example, the file type defines a `force` parameter:
```
nanliu: There should be support for several different property types beyond boolean array to handle whether items are inclusive, does order matter, array of hashes, hashes etc: https://github.com/vmware/vmware-vmware_lib/tree/master/lib/puppet/property

joshcooper: yep, agree it should be there, and work for both parameters and properties.

language/resource_types.md
joshcooper: yep that's mentioned above in https://github.com/puppetlabs/puppet-specifications/pull/72/files/1dd7369b7245c8f4b769a765976408583d426d70#diff-d666e1f4c3db61661d95e6cc0727a82dR102

language/resource_types.md
joshcooper: I don't think that will work, because the catalog is not in scope when evaluating provider features. A provider either statically declares it has that feature:

```ruby
has_feature :manages_symlinks
```

Or the provider feature can be conditional on some other global state like osversion, if dependent gem is present, etc:

```ruby
has_feature :libuser if Puppet.features.libuser?
has_feature :system_groups unless %w{HP-UX Solaris}.include? Facter.value(:operatingsystem)
```

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+Puppet::Type.newtype(:ec2_instance) do
+  # definition goes here
+end
+```
+
+The `newtype` method creates a new class `Puppet::Type::Ec2_instance`, which subclasses `Puppet::Type`. The block passed to the `newtype` method defines methods and variables on the `Puppet::Type::Ec2_instance` class.
+
+## Puppet::Type::&lt;name&gt; methods
+
+Within the `Puppet::Type.newtype` block, the following DSL methods are often called, and are evaluated in the context of the  `Puppet::Type::<name>` class:
```
igalic: "DSL methods" reads very confusing in this context, because all of these methods are called differently (`@foo`, `bar`, `self.baz`). if this is a DSL, it's got some serious abstraction issues.

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+  validate do
+    self.fail "You cannot specify a remote recursion without a source" if !self[:source] && self[:recurse] == :remote
+  end
+end
+```
+
+### autorelations
+
+Allows a type to define automatic relationships (`before`, `subscribe`, `require`, `notify`) between each instance of the type and instances of a different type. For example, the `exec` type will autorequire any file resource whose `path` matches the exec  `command`. Note the relationship will only be added to the catalog if puppet is managing both ends of the relationship. That way puppet will automatically create the file containing the command to execute, before trying to executing it, regardless of manifest order.
+
```
igalic: what? why? why would it do that??

adrienthebo: If you're asking about why autorelations are conditionally created when the autorelated resource is in the catalog, this is because it allows opportunistic relationships to be established. For instance, mounts automatically try to depend on any mounts that are higher up in the file system (https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/type/mount.rb#L287-L294) but don't interrogate the catalog to make sure the other mounts are there. This offloads the complexity from the definition of the autorelations to the actual generation of edges.

The same thing goes for execs - we can opportunistically depend on files that the exec depends on without enforcing that the file is managed, nor do we have to manually interrogate the catalog to make sure the file is there before adding the dependency.

It may be up for debate as to whether this is a good idea but that's probably a lengthy topic in and of itself. If someone has the enthusiasm to discuss this on the dev list and then add support for mandatory autorelationships that would be ace, but for the sake of defining the current behavior I think we're best off focusing on getting the current behavior and rationales documented.

xaque208:
This is great, thank you for putting this together!

language/resource_types.md
joshcooper: Unfortunately the DSL does expose instance variables, instance and class methods. Would it be clearer if I just said that instead of "methods"?

language/resource_types.md
joshcooper: Also `autorequires` predates `manifest ordering`, back when we had [`unguessable_deterministic_key`](https://github.com/puppetlabs/puppet/blob/029dae917ff9b21fbd6388efab7ce8354d9d3d0e/lib/puppet/transaction.rb#L319), which made it impossible to know what order the agent was going to traverse the graph. But then `autorequires` only works for some resources and some of their relationships, which really made things confusing.

That said, I agree with @adrienthebo, we're just looking to define the current behavior in this doc.

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+Defines a new puppet resource type and registers the type using the specified symbolic name. For example, to define a resource type named `ec2_instance`: 
+
+```ruby
+Puppet::Type.newtype(:ec2_instance) do
+  # definition goes here
+end
+```
+
+The `newtype` method creates a new class `Puppet::Type::Ec2_instance`, which subclasses `Puppet::Type`. The block passed to the `newtype` method defines methods and variables on the `Puppet::Type::Ec2_instance` class.
```
MikaelSmith: This seems like a good place to also call out the ability to retrieve existing type classes via `Puppet::Type.type :ec2_instance`.

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+Puppet::Type.newtype(:service) do
+  feature :enableable
+end
+```
+
+And the type specifies that the `enable` property can only be managed if the selected provider supports the `enableable` feature. This validation check is performed at catalog application time, once the provider has been resolved on the agent.
+
+```ruby
+Puppet::Type.newtype(:service) do
+  newproperty(:enable, :required_features => :enableable) do
```
MikaelSmith: Is it possible to define multiple required_features? If so, what's the syntax?

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+        end
+      end
+    end
+  end
+end
+```
+
+## Parameter & Property DSL Methods
+
+The following DSL methods are commonly used for individual parameters and properties within the body of `newparam` and `newproperty` respectively:
```
MikaelSmith: Does that mean there are others you don't mention? I'm not sure how to interpret `commonly` here.

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+
+```ruby
+Puppet::Type.newtype(:ec2_instance) do
+  @doc = 'A type representing an AWS VPC'
+end
+```
+
+### namevar
+
+The type must specify a parameter to be the name of the resource, aka namevar. The name must uniquely identity that resource. By default, a parameter named `name` will be the namevar. This behavior can be modified using `title_patterns` and/or explicitly declaring a parameter to be a namevar.
```
MikaelSmith: Is this a read-only property? Is it callable in some way? Or is it more of a concept? Since you declare this section refers to DSL methods, it's not clear how this fits in.

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+Alternatively, you can pass an option when calling newparam:
+
+```ruby
+Puppet::Type.newtype(:ec2_instance) do
+  newparam(:name, :namevar => true) do
+  end
+end
+```
+
+Note that the namevar is necessarily a parameter, and not a property, since changing the name identifies a different resource as opposed to changing the name of an existing resource.
```
MikaelSmith: It may be worth calling out [package's use](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/package.rb#L268-L272) of namevar. We had a case there that we want to make the provider name part of the namevar. Future changes to the Type API will likely need some way to support this.

language/resource_types.md
joshcooper: Yeah, `Puppet::Type` has 2.6k lines, and some of the code implements the DSL for creating custom types, e.g. `def self.newparam`. But it also implements methods that puppet uses to manage/introspect the type, e.g. `def self.attrclass(name)`, which wouldn't normally be implemented by someone writing a custom type. For example, this might "work" but really shouldn't:

```ruby
Puppet::Type.newtype(:foo) do
  def self.attrclass(name)
    # do terrible things
  end
end
```

I've been focused on writing methods that custom type authors would use, but I may have missed some.

language/resource_types.md
joshcooper: apparently it accepts multiple features...  I'll make a note:

```ruby
Puppet::Type.newtype(:foo) do
  feature :blue, 'blue color'
  feature :green, 'green color'

  newparam(:name) do
  end

  newproperty(:bar, :required_features => [:blue, :green]) do
  end
end
```

language/resource_types.md
joshcooper: sure I can do that

language/resource_types.md
joshcooper: updated

language/resource_types.md
joshcooper: updated

language/resource_types.md
joshcooper: Actually your comment was about the param/property section, so `attrclass` isn't available there, but a similar issue exists for param/properties. But I don't think anything is stopping someone from overriding `def required_features=(*args)` in the body of the `newparam` block.

language/resource_types.md
```diff
@@ -0,0 +1,502 @@
...
+```ruby
+Puppet::Type.newtype(:user) do
+  newproperty(:password) do
+    def is_to_s(currentvalue)
+      return '[old password hash redacted]'
+    end
+  end
+end
+```
+
```
MikaelSmith: For `ensurable`, you mention deferring to the provider's create and destroy methods. Do you also want to call out how properties defer to the provider for `<property>` and `<property>=` as well? Pointed out halfway through https://docs.puppet.com/guides/complete_resource_example.html.

language/resource_types.md
joshcooper: I'll change that to `isnamevar` since that's the actual DSL method. It can also be set in the options hash, e.g. `namevar => true`.

This is a bit in the weeds, but technically the `Puppet::Type#name_var` will return the parameter name (Symbol) or names (Array&lt;Symbol&gt;) that determine the identity for that instance. For example, given:

```ruby
Puppet::Type.newtype(:foo) do
  newparam(:name) do
  end
end
```

Then `Puppet::Type::Foo#name_var` returns:

```
irb(main):011:0> Puppet::Type.type(:foo).new(:name => 'bar').name_var
=> :name
```
That logic will include any parameter whose `Puppet::Parameter.isnamevar?` is truthy or any parameter whose name is `:name`:

```
irb(main):003:0> Puppet::Type.type(:foo).attrclass(:name)
=> Puppet::Type::Foo::ParameterName
irb(main):004:0> Puppet::Type.type(:foo).attrclass(:name).isnamevar?
=> nil
irb(main):005:0> Puppet::Type.type(:foo).key_attribute_parameters
=> [Puppet::Type::Foo::ParameterName]
```

The middle one is surprising because the `isnamevar?` method returns falsey, but puppet effectively treats it as a namevar in [`key_attribute_parameters`](https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/type.rb#L374-L380).

language/resource_types.md
joshcooper: I make a note of the package type in the general "namevar" concept section.

language/resource_types.md
trevor-vaughan: I'd like to say that this is a *very* good idea until you get the ability to do "require (etc...) this if it exists with fancy logic" in the Puppet DSL directly.

It reduces the need for overly complex code and does what users expect. If you want to make it selectable, you can add a parameter that disables auto* or something like that.

language/resource_types.md
trevor-vaughan: How is the catalog not in scope when evaluating provider features? It's happening client side so the catalog should be in scope even if it's stupidly hacky to get at it.

Note: I may be misremembering which part gets evaluated when.

joshcooper: The `has_feature` method is called when the provider is loaded/required in ruby, and that happens on both the agent and server.  In order to be agent-only, the `has_feature` method would need to take a block, and be invoked as part of the `Puppet::Resource#to_ral` process, which converts the server-side resource into an agent-side resource. At that point the catalog would be in-scope, and would behave like `Puppet::Type.validate`, `Puppet::Type.autorequires`.

language/resource_types.md
joshcooper: I thought about not even mentioning how `ensurable` does that, because I wanted to focus on the `Puppet::Type` side of things. I can imagine a separate **provider** doc that talks about how retrieve/insync/sync/mk_resource_methods/etc work.

adrienthebo: +1 for having that documented and +1 for having that in a separate doc. The RAL is a complex creature and there's a lot of things to document, but we don't need to/probabaly shouldn't try to document everything in one shot.

@joshcooper if you would like help documenting the provider API I would be very happy to help.

joshcooper:
I think all comments have been addressed. Please let me know if I've missed something.

hlindberg:
Excellent work and an excellent set of questions. I think this is ready to merge. If something is found to be amiss it can always be updated.
I would love to see the corresponding specification of providers, but I am not sure who has time to devote to that a.t.m.

Bit thanks to everyone involved and @joshcooper for taking the initiative.

PR #73, opened 2016-05-03
(PUP-6173)(PUP-6174) Specify multi-assign from array, hash, and class

hlindberg:
This contains both the missing specification PUP-6173 for multi assignments in general, and for PUP-6174 (new feature) where multiple variables can be accessed "at once" from a class reference.

### DIFF
--- a/language/expressions.md
+++ b/language/expressions.md
@@ -709,7 +709,7 @@ An L-value is a name referring to a named "slot" in the current scope (that is, 
   * *Array* - each variable is assigned from the RHS array's index 0-n. It is an error if there are too few values.
   * *Hash* - each variable is assigned the value from the corresponding key in the RHS hash. It is in an error if the key is not present.
   * *Type[Class]* - each variable is assigned the value of the corresponding variable/parameter in the referenced class. The
-    class must have been included, and the referenced variable must exist, but may have an `undef` value assigned.
+    class must have been added to the catalog, and the referenced variable must exist, but may have an `undef` value assigned.
 
 Assignment also takes place in parameter declarations of user defined resource types and
 classes. An alternate form of assignment also takes place when resource attributes are set.
--- a/language/expressions.md
+++ b/language/expressions.md
@@ -699,7 +699,7 @@ The assignment operator assigns the result of the RHS to one or more L-values pr
 An L-value is a name referring to a named "slot" in the current scope (that is, typically a variable).
 
 * A `$` variable produces an L-value name
-* Only a Simple Name is accepted, it is not allowed to assign to something in another namepace
+* Only a Simple Name is accepted, it is not allowed to assign to something in another namespace
 * Numerical L-values are not allowed (numerical variables are read-only and set by side effect
   of matching with a regular expression).
 * Assignment is an R-value
--- a/language/expressions.md
+++ b/language/expressions.md
@@ -695,15 +695,21 @@ The following table shows the result of searching for a LHS of a particular type
 Assignment Operator
 ---
 
-The assignment operator assigns the result of the RHS to the L-Value produced by the LHS. An L-value is a name referring to a "slot" in the current scope (that can be referenced (typically by a 
-variable) to obtain the value).
+The assignment operator assigns the result of the RHS to one or more L-values produced by the LHS expression.
+An L-value is a name referring to a named "slot" in the current scope (that is, typically a variable).
 
 * A `$` variable produces an L-value name
-* Only a Simple Name is accepted
+* Only a Simple Name is accepted, it is not allowed to assign to something in another namepace
 * Numerical L-values are not allowed (numerical variables are read-only and set by side effect
   of matching with a regular expression).
 * Assignment is an R-value
 * The value of an assignment is the value of the RHS
+* An array of L-values forms a multi-assignment where several variables can be assigned at once from the given RHS source
+* In a multi-assignment the result depends on the type of the RHS:
+  * *Array* - each variable is assigned from the RHS array's index 0-n. It is an error if there are too few values.
+  * *Hash* - each variable is assigned the value from the corresponding key in the RHS hash. It is in an error if the key is not present.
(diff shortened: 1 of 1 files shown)


language/expressions.md
```diff
@@ -695,15 +695,21 @@ The following table shows the result of searching for a LHS of a particular type
...
 * Numerical L-values are not allowed (numerical variables are read-only and set by side effect
   of matching with a regular expression).
 * Assignment is an R-value
 * The value of an assignment is the value of the RHS
+* An array of L-values forms a multi-assignment where several variables can be assigned at once from the given RHS source
+* In a multi-assignment the result depends on the type of the RHS:
+  * *Array* - each variable is assigned from the RHS array's index 0-n. It is an error if there are too few values.
+  * *Hash* - each variable is assigned the value from the corresponding key in the RHS hash. It is in an error if the key is not present.
+  * *Type[Class]* - each variable is assigned the value of the corresponding variable/parameter in the referenced class. The
+    class must have been included, and the referenced variable must exist, but may have an `undef` value assigned.
```
ripienaar: would `parsed` be better? since presumably `inherits` would do the same and it doesn't really cover the resource like syntax?

language/expressions.md
```diff
@@ -720,9 +726,26 @@ result. Chained assignments are permitted.
...
+
+    # from hash
+    [$a, $b] = { a => 10, b => 20, c => 30 }  # assigns 10 to $a, and 20 to $b
+
+    # from class
+    class mymodule::someclass::example($x = 100) {
+      $a = 10
+    }
+    include example
+    [$a, $x] = Class['mymodule::someclass::example']  # assigns 10 to $a, and 100 to $x
```
ripienaar: surprised me that this is fetching *all* variables in the class and not just parameters, kind of leaky about the internals of a class? 

Of course you could just fetch `$mymodule::someclass::example::a` so it's no worse than before.

hlindberg: Unfortunately (by design) all variables in a class are externally available and thus part of the class' API. Even more unfortunate is that the set of variables can be conditional.
This does not alter the old design decisions though. Instead, this is intended to help in situations where something private/implementation specific should be exposed as a set of documented and supported variables in a modules "API classes". You can do the same now by just typing a lot. The interesting thing you can now do (which was horrible to do in the past), is to assign variables from a selection of class based on a  variable - for example [$a, $x] = Class[$the_inner_class_used], which allows an advanced module that configures itself to easily choose which variables to expose - result is a lot clearer than lots of lines of variable assignment intermixed with a bit of if-then-else logic.

ripienaar: Yeah, I see the use for sure, final docs should probably just be clear to call out that this will export all variables including private ones.

And I would think something like this might be handy in wrapper class type situation where you'd extract all the expected parameters from another class or define, except you can't because you also get internal stuff?

Are you looking to add this to resources as well, including defines?

language/expressions.md
hlindberg: "parsed" is something completely different - both the resource instantiation expression and a call to include - includes the class in the catalog. The documentation uses the term "declared" (a term that I strongly dislike for this concept).

I can change to the more colloquial "added to the catalog" if that helps?

language/expressions.md
hlindberg: You cannot extract from a define, only from classes. No plan to add it to defines because they are evaluation order dependant, and it would only be their parameters that can be accessed. Also, the final values of resource parameters are in some cases lazily determined (because of reource defaults expressions, and overrides). So - no, too much whimsical old behaviour involved there.

language/expressions.md
ripienaar: yes, `declared` is horrible, I don't know what the right term is, I also use `included` for this purpose when I talk about it, but some people get confused.  `added` might work yes.

language/expressions.md
ripienaar: hmm, so

```
define x($y) { }
x{"meh": y => 1}
notice(X["meh"]["y"])
```

works, seems you should be able to get at the parameters at least?

As you say its evaluation order dependant, but I thought so was the class case since the internal vars wouldnt be available if not included already?

hlindberg: A class added to the catalog using `class { classname: }` is lazily evaluated. All other - via include, are available immediately. There is really no such thing as a "private" variable in a class (except those that start with an underscore - since they are not valid in fully qualified name form. 
Your example works because both your instantiation of "meh" and the "notice" are lazily evaluated (and the notice is behind the "meh" in the queue...

ripienaar: ok, thanks for explaining

hlindberg: You can get parameters from a define with `X[meh][x]`  but only if it has been evaluated. We could add support for multi-assign for that, but people will probably be bit by evaluation order all the time trying to use that. Also, since the parameters to the defines must come from somewhere - and that is usually from classes; seems better to get them from there (i.e. from the input, rather from the "resulting resource").

language/expressions.md
hlindberg: new commit pushed onto this PR

language/expressions.md
```diff
@@ -695,15 +695,21 @@ The following table shows the result of searching for a LHS of a particular type
...
 ---
 
-The assignment operator assigns the result of the RHS to the L-Value produced by the LHS. An L-value is a name referring to a "slot" in the current scope (that can be referenced (typically by a 
-variable) to obtain the value).
+The assignment operator assigns the result of the RHS to one or more L-values produced by the LHS expression.
+An L-value is a name referring to a named "slot" in the current scope (that is, typically a variable).
 
 * A `$` variable produces an L-value name
-* Only a Simple Name is accepted
+* Only a Simple Name is accepted, it is not allowed to assign to something in another namepace
```
thallgren: Typo 'namepace'

hlindberg: "The rate at which names change?"(fixing)

hlindberg:
Merging since there was only one typo to fix from review.

PR #74, opened 2016-05-04
(PUP-5966) Specify new operation

hlindberg:
This adds a specification of the new operation:
* Ability to call a type (general information)
* Specification per type that supports new

### DIFF
--- a/language/types_values_variables.md
+++ b/language/types_values_variables.md
@@ -490,6 +490,237 @@ only, it can not be used in the Puppet Programming Language.
     String ∪ (T ∈ Scalar)  → Scalar
     String ∪ (T ∉ Scalar)  → Any
 
+#### String.new
+
+Since version 4.5.0
+
+Conversion to String is the most comprehensive conversion as there are many
+use cases where a String representation is wanted. The defaults for the many options
+have been chosen with care to be the most basic "value in textual form" representation.
+
+A new String can be created from all other data types. The process is performed in
+several steps - first the type of the given value is inferred, then the resulting type
+is used to find the most significant format specified for that type. And finally,
+the found format is used to convert the given value.
+
+The mapping from type to format is referred to as the format map. This map
+allows different formatting depending on type.
+
+Example Positive Integers in Hexadecimal prefixed with '0x', negative in Decimal
+
+~~~ puppet
+
+$format_map = { Integer[default, 0] => "%d",
+  Integer[1, default] => "%#x"
+}
+String("-1", $format_map)  # produces '-1'
+String("10", $format_map)  # produces '0xa'
+
+~~~
+
+A format is specified on the form:
+
+~~~
+
+%[Flags][Width][.Precision]Format
+
+~~~
+
+`Width` is the number of characters into which the value should be fitted. This allocated space is
+padded if value is shorter. By default it is space padded, and the flag 0 will cause padding with 0
+for numerical formats.
+
+`Precision` is the number of fractional digits to show for floating point, and the maximum characters
+included in a string format.
+
+Note that all data type supports the formats `s` and `p` with the meaning "default to string" and
+"default programmatic string representation" (which for example means that a String is quoted in 'p' format).
+
+##### Signatures of String conversion
+
+~~~ puppet
+
+type Format = Pattern[/^%([\s\+\-#0\[\{<\(\|]*)([1-9][0-9]*)?(?:\.([0-9]+))?([a-zA-Z])/]
+type ContainerFormat = Struct[{
+  format         => Optional[String],
+  separator      => Optional[String],
+  separator2     => Optional[String],
+  string_formats => Hash[Type, Format]
+  }]
+type TypeMap = Hash[Type, Variant[Format, ContainerFormat]]
+type Formats = Variant[Default, String[1], TypeMap]
+
+Callable[Any, Formats]
+
+~~~
+
+Where:
+* `separator` is the string used to separate entries in an array, or hash (extra space should not be included at
+  then end), defaults to `","`
+* `separator2` is the separator between key and value in a hash entry (space padding should be included as
+  wanted), defaults to `" => ".
+* `string_formats` is a type to format map for values contained in arrays and hashes - defaults to `{Any => "%p"}`. Note that
+  these nested formats are not applicable to containers which are always formatted as per the top level format specification.
+
+Example Simple Conversion to String (using defaults)
+
+~~~ puppet
+
+$str = String(10)      # produces '10'
+$str = String([10])    # produces '["10"]'
+
+~~~
+
+Example Simple Conversion to String specifying the format for the given value directly
+
+~~~ puppet
+
+$str = String(10, "%#x")    # produces '0x10'
+$str = String([10], "%(a")  # produces '("10")'
+
+~~~
+
+Example Specifying type for values contained in an array
+
+~~~ puppet
+
+$formats = { Array => {format => '%(a', string_formats => { Integer => '%#x' } }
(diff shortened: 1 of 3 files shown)


language/types_values_variables.md
```diff
@@ -337,6 +366,52 @@ Iterating over an integer range:
...
+* When converting from `String` the default radix is 10
+* If radix is not specified an attempt is made to detect the radix from the start of the string:
+  * `0b` or `0B` is taken as radix 2
+  * `0x` or `0X` is taken as radix 16 
+  * `0` as radix 8
+  * all other are decimal
+* Conversion from `String` accepts an optional sign in the string.
+* For hexadecimal (radix 16) conversion an optional leading "0x", or "0X" is accepted
+* for octal (radix 8) an optional leading "0" is accepted
+* for binary (radix 2) an optional leading "0b" or "0B" is accepted
```
thallgren: The above three lines could be condensed into one line saying something like:
"An optional radix prefix may be combined with a radix parameter if both denote the same radix".

language/types_values_variables.md
```diff
@@ -337,6 +366,52 @@ Iterating over an integer range:
...
+* When `radix` is set to `default`, the conversion is based on the leading
+  characters in the string. A leading "0" for radix 8, a leading "0x", or "0X" for
+  radix 16, and leading "0b" or "0B" for binary.
```
thallgren: Unnecessary to repeat explanation of the prefix.

language/types_values_variables.md
```diff
@@ -337,6 +366,52 @@ Iterating over an integer range:
...
+* Conversion from `Boolean` results in 0 for `false` and 1 for `true`.
+* Conversion from `Integer`, `Float`, and `Boolean` ignores the radix.
```
thallgren: I'd prefer: "Radix is only applicable to `String` conversion (ignored for other types)" (better to state what works than all combinations that doesn't).

language/types_values_variables.md
```diff
@@ -337,6 +366,52 @@ Iterating over an integer range:
...
+* `Float` value fractions are truncated (no rounding)
+
+Example Converting to Integer
+
+~~~ puppet
+
+$a_number = Integer("0xFF", 16)  # results in 255
+$a_number = Numeric("010")       # results in 8
+$a_number = Numeric("010", 10)   # results in 10
```
thallgren: Perhaps add:
```
$a_number = Numeric("0x10", 10)   # this is an error. Prefix and radix does not match
```

language/types_values_variables.md
```diff
@@ -362,6 +437,18 @@ You can learn more about floating point than you ever want to know from these ar
     Float ∪ (T ∉ Scalar)        → Any
     Float[a, b] ∪ Float[c, d]   → Float[min(a, c), max(b,d)]
 
+#### Float.new
+
+Since version 4.5.0
+
+A new `Float` can be created from `Integer`, `Float`, `Boolean`, and `String` values.
+For conversion `from` String both float and integer formats are supported.
```
thallgren: Don't think you intended to quote the word `from` here. `String` perhaps?

thallgren: should have comma after String

language/types_values_variables.md
```diff
@@ -415,6 +502,237 @@ only, it can not be used in the Puppet Programming Language.
...
+
+A new String can be created from all other data types. The process is performed in
+several steps - first the type of the given value is inferred, then the resulting type
+is used to find the most significant format specified for that type. And finally,
+the found format is used to convert the given value.
+
+The mapping from type to format is referred to as the format map. This map
+allows different formatting depending on type.
+
+Example Positive Integers in Hexadecimal prefixed with '0x', negative in Decimal
```
thallgren: Colon after Example.

language/types_values_variables.md
```diff
@@ -415,6 +502,237 @@ only, it can not be used in the Puppet Programming Language.
...
+
+~~~ puppet
+
+$format_map = { Integer[default, 0] => "%d",
+  Integer[1, default] => "%#x"
+}
```
thallgren: weird indentation of map. Hard to read.

thallgren: Also, I don't think 0 is a negative integer ;-)

language/types_values_variables.md
```diff
@@ -415,6 +502,237 @@ only, it can not be used in the Puppet Programming Language.
...
+
+`Width` is the number of characters into which the value should be fitted. This allocated space is
+padded if value is shorter. By default it is space padded, and the flag 0 will cause padding with 0
+for numerical formats.
+
+`Precision` is the number of fractional digits to show for floating point, and the maximum characters
+included in a string format.
+
+Note that all data type supports the formats `s` and `p` with the meaning "default to string" and
+"default programmatic string representation" (which for example means that a String is quoted in 'p' format).
```
thallgren: "default string representation" and "default programmatic string representation" is more consistent than using "default to string".

language/types_values_variables.md
```diff
@@ -415,6 +502,237 @@ only, it can not be used in the Puppet Programming Language.
...
+Given formats are merged with the default formats, and matching of values to convert against format is based on
+the specificity of the mapped type; for example, different formats can be used for short and long arrays.
+
+##### Integer to String
+
+| Format  | Integer Formats
+| ------  | ---------------
+| d       | Decimal, negative values produces leading '-'
+| x X     | Hexadecimal in lower or upper case. Uses ..f/..F for negative values unless # is also used
+| o       | Octal. Uses ..0 for negative values unless # is also used
```
thallgren: I don't understand what "..f/..F for negative values means".

language/types_values_variables.md
```diff
@@ -415,6 +502,237 @@ only, it can not be used in the Puppet Programming Language.
...
+| b B     | Binary with prefix 'b' or 'B'. Uses ..1/..1 for negative values unless # is also used
```
thallgren: I don't understand what "..1/..1 for negative values" means

language/types_values_variables.md
hlindberg: While I do agree it could be shortened it requires reformulating the entire thing as the prefixes and radix values are not defined as it is not written.

language/types_values_variables.md
hlindberg: While I do agree it could be shortened it requires reformulating the entire thing as the prefixes and radix values are not defined as it is not written.

language/types_values_variables.md
hlindberg: ok, fixing

language/types_values_variables.md
hlindberg: ok, fixing

language/types_values_variables.md
hlindberg: doh... fixing

language/types_values_variables.md
hlindberg: ok, fixing

language/types_values_variables.md
hlindberg: merde... fixing

language/types_values_variables.md
hlindberg: The "..f" or "..F" indicates an infinite series of "F" since the value is negative, and the size of the integer is unspecified.  Can add an explanation for that (although there is not much room in the table (hard to write lots of text in a cell without using HTML tables).

language/types_values_variables.md
hlindberg: 1 in infinity since the size of the integer is not specified.

hlindberg:
All comments addressed. Merging.

PR #78, opened 2016-07-05
(maint) Fix how puppetdb is referenced in file path doc

briancain:
I know at least the product should be referenced as `puppetdb` instead of `puppet-db`, but does anyone know if the conf file is `puppetdb.conf` vs `puppet-db.conf` ? @ajroetker @wkalt


ajroetker:
It's interesting that this one was already correct https://github.com/puppetlabs/puppet-specifications/pull/78/files#diff-1b625f7fd389243fcb8401c1dfb33b00L352

@briancain puppetdb and puppetdb.conf all the way thanks for the PR/ping. 👍  from me

briancain:
@ajroetker huh yeah strange...must of just been a mistake when the hyphen was added.

Who typically merges stuff on this repo... @hlindberg perhaps?

kylog:
Thanks @briancain and @ajroetker!

PR #81, opened 2016-08-12
(PUP-6599) Specify return type for function, lambda and Callable

hlindberg:


language/puppet-functions.md
```diff
@@ -7,16 +7,20 @@ Grammar
 ---
 
     FunctionDefinition
-      : 'function' parameters = ParameterList? '{' statements += statements '}'
+      : 'function' name = NAME parameters = ParameterList? ReturnType? '{' statements += statements '}'
```
thallgren: Should this read:
```
      : 'function' name = NAME parameters = ParameterList? return_type = ReturnType? '{' statements += statements '}'
```

hlindberg: does not really matter, only entries where the name is needed in the text should have an assignment (or where it acts as a clarification of what the value means). Here it is redundant as the rule is already named "ReturnType" and "return_type = ReturnType" does not give any new information. OTOH, "name = NAME" is equally redundant, I can remove that to make it shorter/more readable.

thallgren: My only concern here is consistency. People not familiar with this type of syntax will find formal description harder to read when the syntax varies depending on what's included in the text or not.

hlindberg: Agree, it does however vary throughout the entire specification suite (names/assignments are only used where needed). IMO, names should only be used when they add something - they otherwise only makes it harder to read

PR #84, opened 2016-09-06
(PUP-5871) Add specification of Timestamp and Timespan

thallgren:


language/expressions.md
```diff
@@ -1385,6 +1474,98 @@ Examples:
...
+* The range values may be negative.
+
+Examples:
+
+    Timespan[2]                        # a timespan of 2 seconds
+    Timespan[77.3]                     # a timespan of 1 minute, 17 seconds, and 300 milliseconds
+    Timespan[{hour => 1}, {hour => 2}] # a timespan between 1 and two hours
+    Timespan['1-00:00:00', '2-00:00:00'] # a timespan between 1 and 2 days
+    Timespan[Timespan('11', '%H'), Timespan('12', '%H')]              # a timespan between 11 and 12 hours
+    Timespan['1-00:00:00', Timespan({days => 2, nano-seconds => -1})] # a timespan between 1 and up to (but not including) 2 days
```
hlindberg: cannot use `nano-seconds` as a hash key, it ends up being a subtraction of strings "seconds" from "nano". Needs quoting.

kylog: How about just `nanoseconds`? I've always seen it treated as a single word, e.g. at https://en.wikipedia.org/wiki/Nanosecond.

language/expressions.md
```diff
@@ -1385,6 +1474,98 @@ Examples:
...
+* A value of default means -Infinity if given as the value of `from`, and +Infinity if given
+  as `to`. If only an exact value of `default` is given this is the same as
+  `Timestamp[default, default]`,
+  which again is the same as just `Timestamp` (i.e. all floating point values in the range +/- Infinity).
+* The `from` value may be greater than the `to` value.
+* The range values may be negative.
+
+Examples:
+
+    Timestamp['2000-01-01T00:00:00.000', :default] # a timestamp in the 21st century or later
```
hlindberg: remove the ':' - not applicable in puppet language

language/expressions.md
```diff
@@ -1385,6 +1474,98 @@ Examples:
...
+    Timestamp['1-00:00:00', '2-00:00:00'] # a timespan between 1 and 2 days
+    Timestamp[Timestamp('11', '%H'), Timestamp('12', '%H')]              # a timespan between 11 and 12 hours
+    Timestamp['1-00:00:00', Timestamp({days => 2, nano-seconds => -1})] # a timespan between 1 and up to (but not including) 2 days
```
hlindberg: hyphen in `nano-seconds` is not correct

kylog:
Oh I see I commented on just one line wrt `nanosecond` vs `nano-second`.

I would treat `milliseconds`, `microseconds` and `nanoseconds` as single words throughout. E.g. see:
* https://en.wikipedia.org/wiki/Microsecond
* https://en.wikipedia.org/wiki/Millisecond
* https://en.wikipedia.org/wiki/Nanosecond

Thoughts?

language/types_values_variables.md
```diff
@@ -142,16 +146,17 @@ may appear more than once in the hierarchy (e.g. a `Scalar` is both `Any` and `D
        |  |     |- (Integer with range inside another Integer)
        |  |  |- Float[from, to]
        |  |     |- (Float with range inside another Float)
+       |  |  |- Timespan
+       |  |  |- Timestamp
```
hlindberg: This looks problematic, while they can be involved in arithmetic they are not numeric data types in terms of Data (basically something that can be expressed in Json with only Json data types). There will also be problems if a type accepts Numeric and is not prepared to handle TimeStamp or TimeSpan. Arithmetic rules are also not cumulative. Suggest that they are not considered to be of Numeric type.

language/types_values_variables.md
```diff
@@ -142,16 +146,17 @@ may appear more than once in the hierarchy (e.g. a `Scalar` is both `Any` and `D
...
        |  |  |- Enum[*strings]
        |  |  |- Pattern[*patterns]
        |  |
        |  |- Boolean
        |  |- Regexp[pattern_string]
-       |
-       |- SemVer
-       |- SemVerRange
+       |  |- SemVer
+       |  |- SemVerRange
```
hlindberg: This is problematic because Scalar is considered Data IIRC. We then end up with a translation problem as SemVer and SemVerRange cannot be represented as Data without using some kind of mapping. (That was the reason they were not added to the table as Scalars even if they are scalars in practice). It makes more sense to fix Scalar in Data then not having SemVer types be Scalar though.

language/types_values_variables.md
```diff
@@ -269,7 +274,7 @@ Represents the abstract notion of "value", its subtypes are `Numeric`, `String`
 
 ### Numeric
 
-Represents the abstract notion of "number", its subtypes are `Integer`, and `Float`.
+Represents the abstract notion of "number", its subtypes are `Integer`, `Float`, `Timespan` and `Timestamp`.
```
hlindberg: I think this is problematic as stated elsewhere. We would need a type that is Variant[Integer, Float] since that is what Numeric is now. We could add a "NumericBasedScalar" or similar of which Numeric is a sub type, but I am not sure that adds any value compared to just making Timespan and Timestamp be Scalars since other numeric based scalars would probably have other kinds of constraints in the arithmetic contract.

language/types_values_variables.md
```diff
@@ -456,6 +461,333 @@ For conversion from `String` both float and integer formats are supported.
...
+
+```puppet
+function Timespan.new(
+  Variant[Float, Integer] $value
+)
+```
+
+##### Timespan from days, hours, mintues, seconds, and fractions of a second
+
+The arguments can be passed separately in which case the first four, days, hours, minutes, and seconds are mandatory and the rest are optional.
```
hlindberg: The `first four;` (now it looks like the word `four` is part of the enumeration, use a semicolon to make the enumeration an explanation of what "the first four" means).

language/types_values_variables.md
```diff
@@ -456,6 +461,333 @@ For conversion from `String` both float and integer formats are supported.
...
+)
+```
+
+##### Timespan from String and format directive patterns
+
+The first argument is parsed using the format optionally passed as a string or array of strings. When an array is used, an attempt
+will be made to parse the string using the first entry and then with each entry in succession until parsing succeeds. If the second
+argument is omitted, an array of default formats will be used.
+
+An exception is raised when no format was able to parse the given string.
```
hlindberg: We do not raise exception in puppet - the concept of exception does not exist. We do "raise errors". (If other text in the spec talks about exceptions it should be changed). Or, since this is a specification and the mechanism is not specified - it can simply say "It is an error if no format was able to..."

language/types_values_variables.md
```diff
@@ -456,6 +461,333 @@ For conversion from `String` both float and integer formats are supported.
...
+)
+```
+
+##### Timestamp from String and patterns consisting of format directives
+
+The first argument is parsed using the format optionally passed as a string or array of strings. When an array is used, an attempt
+will be made to parse the string using the first entry and then with each entry in succession until parsing succeeds. If the second
+argument is omitted, an array of default formats will be used.
+
+An exception is raised when no format was able to parse the given string.
```
hlindberg: Use of "exception" (and I checked, we do not use the term in the specification).

language/types_values_variables.md
```diff
@@ -646,6 +978,28 @@ as negative value.
 
 Defaults to `p`
 
+##### Timespan to String
+
+| Format  | Timespan formats
+| ----    | ------------------
+| s       | formats according to the timespan format string '%D-%H:%M:%S' 
+| p       | programmatic representation - a quoted string formatted according to the timespan format string '%D-%H:%M:%S'
```
hlindberg: Hm, that is not an accurate programmatic representation - it would need to say Timestamp('...'), or Timespan('...') for that to be true since you cannot parse the value back again to the same data type.

language/expressions.md
hlindberg: +1 to that change

language/types_values_variables.md
```diff
@@ -456,6 +461,333 @@ For conversion from `String` both float and integer formats are supported.
...
+All values may overflow and/or be negative. The internal 128-bit nano-second integer is calculated as:
+
+```
+(((((days * 24 + hours) * 60 + minutes) * 60 + seconds) * 1000 + milli_seconds) * 1000 + micro_seconds) * 1000 + nano_seconds
+```
+
+```puppet
+function Timespan.new(
+  Integer $days, Integer $hours, Integer $minutes, Integer $seconds,
+  Integer $milli_seconds = 0, Integer $micro_seconds = 0, Integer $nano_seconds = 0
```
hlindberg: So, per @kylog 's suggestion, these would need to change to not use underscore

language/types_values_variables.md
```diff
@@ -456,6 +461,333 @@ For conversion from `String` both float and integer formats are supported.
...
+function Timespan.new(
+  Struct[{
+    Optional[negative] => Boolean,
+    Optional[days] => Integer,
+    Optional[hours] => Integer,
+    Optional[minutes] => Integer,
+    Optional[seconds] => Integer,
+    Optional[milli_seconds] => Integer,
+    Optional[micro_seconds] => Integer,
+    Optional[nano_seconds] => Integer
```
hlindberg: Same here (underscore).

language/types_values_variables.md
hlindberg: "exception"

language/expressions.md
thallgren: If nanosecond is one word, then I assume that we should also use microsecond and millisecond?

language/types_values_variables.md
thallgren: I don't understand. Why is a `SemVer` different from a `String` in this matter? To me, both are scalars and data. I have the same question regarding `SemVerRange` and `Regexp`.

language/types_values_variables.md
thallgren: The implementation is based on the fact that they are numeric and represents seconds. I'm not sure I see the problem in viewing them that way.

language/types_values_variables.md
thallgren: Perhaps it's better to change this so that 'p' is the same as 's' ?

hlindberg: The best would be to output Timestamp('str') and Timespan('str') for %p format.

language/expressions.md
hlindberg: yes

language/types_values_variables.md
hlindberg: Regexp is problematic. My mistake.  Since we have that problem, and the constraints of Data are not really used for what it was intended for, I don't think it matters much if SemVer and SemVerRange are of Scalar type.
The problem is that Regexp, SemVer, etc. cannot be trivially represented as data directly represented in JSon (you would need to add interpretation).

language/types_values_variables.md
hlindberg: They break the contract for Numeric. Someone accepting Numeric may get values that will cause errors as all arithmetic operations are not applicable using the Numeric contract. They also represent values you cannot have in a catalog (yet). They are therefore not Numeric in the same sense even if they can be coerced to a representation that is. If they are truly numeric, then illogical operations like adding two Timestamps must be allowed.

language/types_values_variables.md
thallgren: I guess we should add similar entries for SemVer('str') and SemVerRange('str').  I think the String conversion of those types is missing altogether.

language/types_values_variables.md
thallgren: Changing this isn't just confined to the spec. It will break the implementation as well. Things like common type between a Float and Timespan for instance. It can no longer be Numeric.

language/types_values_variables.md
thallgren: Oh, I didn't realize that `Data` was constrained to json data types. That would rule out `Pattern` and `Regexp` as well. If that's what it is, then why not remove `Scalar` from the tree beneath `Data` and instead add `Integer`, `Float`, `String`, and `Boolean`? A `Scalar` will then not be constrained to json.

language/types_values_variables.md
hlindberg: True, the implementation should change.

language/types_values_variables.md
hlindberg: Yes, sounds sensible. If a user want the actual String, they can switch to using the %s format.

language/types_values_variables.md
thallgren: But isn't this a bug today? If `Data` is confined to JSON data types, then it _should_ break when a `Regexp` is used.

thallgren: Also, now we make the problem worse by introducing `SemVer`, `SemVerRange`, `Timespan`, and `Timestamp` as `Scalar` values.

thallgren: I would expect that we, in a near future, also add `URI` and perhaps `UUID` to the set of `Scalar` values.

thallgren:
Superseded by [PR-85](/puppetlabs/puppet-specifications/pull/85).

language/types_values_variables.md
hlindberg: It is a bug that Regexp is Data. It has not been fixed yet because it is a breaking change. Not sure we have a ticket for that.

PR #85, opened 2016-09-08
(PUP-6688) Add specification of Timestamp and Timespan.

thallgren:
Supersedes [PR-84](/puppetlabs/puppet-specifications/pull/84).

### DIFF
--- a/language/types_values_variables.md
+++ b/language/types_values_variables.md
@@ -50,10 +50,21 @@ the conceptual categories **Data Types** e.g.:
 *  `Integer`
 *  `Float`
 *  `Boolean`
-*  `Regexp`
 *  `String`
 *  `Array`
 *  `Hash`
+*  `Undef`
+
+**Scalar Types** e.g.:
+*  `Integer`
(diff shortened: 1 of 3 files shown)


language/types_values_variables.md
```diff
@@ -646,6 +986,42 @@ as negative value.
 
 Defaults to `p`
 
+##### Timespan to String
+
+| Format  | Timespan formats
+| ----    | ------------------
+| s       | formats according to the timespan format string '%D-%H:%M:%S' 
+| p       | programmatic representation - "Timespan(<quoted string>)" where <quoted string> is the result of using '%s'
```
hlindberg: Does %s produce a quoted String? - i.e. it is a bit unclear if "Timespan(12345)" or "Timespan('12345')" is produced for %p. I thought %s would produce an unquoted string.
(The same comment applies to Timestamp below).

thallgren: Yes, it should say "is the quoted result of using '%s'", changing.

PR #89, opened 2016-10-14
(PE-10715) Move mcollective logs into their own directory

npwalker:
Prior to this commit, mcollective logs were placed directly in
/var/log/puppetlabs which is the only component to not have
its own directory

After this commit, mcollective logs go in a mcollective directory


MikaelSmith:
What kind of breaks should we be concerned about from changing this? If anyone has watchers set on the file, they'll need to update them. I think I'm ok doing that in a feature release and shouting it loudly.

PR #90, opened 2016-11-03
(PCP-458) Add pcp-access.log file path

MikaelSmith:
The next version of pxp-agent will add an access log that notes all
messages received. For details, see PCP-387. This commit adds the
default locations of those access logs.

This depends on https://github.com/puppetlabs/pxp-agent/pull/477 being merged.


MikaelSmith:
pxp-agent changes merged.

PR #91, opened 2016-11-18
(PUP-6934) Update file path specification to include locale directories

Magisus:
This adds a specification for the location of translation files for use
with gettext, both on Unix and on Windows.

This PR is in part a request for comment: do these seem like the correct directories for PO files?

### DIFF
--- a/file_paths.md
+++ b/file_paths.md
@@ -136,6 +136,7 @@ The files annotated by an '*' indicate that they are created by package installa
                     LC_MESSAGES
                         <project>.mo *    # One directory for each supported locale,
                                           # containing the MO files for each project.
+                                          # Example for English: /opt/puppetlabs/puppet/share/locale/en_US/LC_MESSAGES/puppet.mo
         ssl *
         VERSION                           # puppet-agent package version
 
@@ -302,6 +303,7 @@ create a `puppet` user or group.
                 LC_MESSAGES
                     <project>.mo              # One directory for each supported locale,
                                               # containing the MO files for each project.
+                                              # Example for English: C:\Program Files\Puppet Labs\Puppet\puppet\locale\en_US\LC_MESSAGES\puppet.mo
 
     C:\Program Files\Puppet Labs\Puppet\pxp-agent *
         bin *
--- a/file_paths.md
+++ b/file_paths.md
@@ -298,6 +298,10 @@ create a `puppet` user or group.
         lib *
             puppet.rb *
         locale *
+            <lang_COUNTRY>
+                LC_MESSAGES
+                    <project>.mo              # One directory for each supported locale,
+                                              # containing the MO files for each project.
 
     C:\Program Files\Puppet Labs\Puppet\pxp-agent *
         bin *
--- a/file_paths.md
+++ b/file_paths.md
@@ -131,6 +131,11 @@ The files annotated by an '*' indicate that they are created by package installa
             augeas *
             man *
             vim *
+            locale
+                <lang_COUNTRY>
+                    LC_MESSAGES
+                        <project>.mo *    # One directory for each supported locale,
+                                          # containing the MO files for each project.
         ssl *
         VERSION                           # puppet-agent package version
 
@@ -292,6 +297,7 @@ create a `puppet` user or group.
             puppet *                          # ruby bin wrapper
         lib *
             puppet.rb *
+        locale *
 
     C:\Program Files\Puppet Labs\Puppet\pxp-agent *
         bin *
--- a/file_paths.md
+++ b/file_paths.md
@@ -131,12 +131,14 @@ The files annotated by an '*' indicate that they are created by package installa
             augeas *
             man *
             vim *
-            locale
-                <lang_COUNTRY>
-                    LC_MESSAGES
+            locale *
+                config.yaml *             # configuration file used by the gettext-setup gem
+                <lang_COUNTRY> *
+                    LC_MESSAGES *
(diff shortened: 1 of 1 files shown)


glennsarti:
👍 for merge from me

file_paths.md
```diff
@@ -292,6 +294,7 @@ create a `puppet` user or group.
             puppet *                          # ruby bin wrapper
         lib *
             puppet.rb *
+        locale *
```
Iristyle: So on Windows we have the siblings:
* `bin`
* `lib`
* `locale`

But on the *nix variant above, we have siblings like:

* `bin`
* `lib`
* `share`
  * `puppet`
    * `locale`

So the two relative path structures are not the same.

I see references to a top-level `/opt/share/locale` ... so wondering if `/opt/puppetlabs/puppet/locale` is a sufficient starting point? (i.e. what advantage is there to nesting under `share`)?  That's probably a question for @branan

Iristyle: I also see that if we call `msgfmt` - I see a reference in the i18n build workflow doc at https://davesteele.github.io/debian/development/2015/12/10/i18n-build-flow/ to put them in `/usr/share/locale/<language>[-<COUNTRY>]/LC_MESSAGES/<package name>.mo`

But I don't know if there's any reason for that, etc.

Iristyle: Might be good to ask @lutter / @scottyw / @LarissaLane here, because I'm just coming up to speed on how the translation workflow actually works.

LarissaLane: From the translation side, I specify the location of the PO file using the file path like this:

`[transifex-test-3.japo]
source_file = locales/ja_JP/ja.po
source_lang = en
type = PO
lang_map = ja: ja_JP
file_filter = locales/ja_JP/ja.po` 

I don't think it matters on the Transifex side where the PO file is located, as long as we provide the correct file path.

Magisus:
Some context from further research:

Unless I'm mistaken, it looks like the C++ tooling is going to require a specific directory structure (see http://www.boost.org/doc/libs/1_62_0/libs/locale/doc/html/messages_formatting.html#msg_loading_dictionaries). Do we want this to affect the way we structure things for Ruby at all? Alternatively, am I missing something about our C++ process that makes this irrelevant? /cc @MikaelSmith

Magisus:
Updated this to use something more standard on Unix:

`/opt/puppetlabs/puppet/share/locale/<lang_COUNTRY>/LC_MESSAGES/<project>.mo`

Magisus:
If we want to use the Unix standard path structure described above, we will have to make changes to  the gettext-setup gem, because it is currently [hard-coded](https://github.com/puppetlabs/gettext-setup-gem/blob/master/lib/gettext-setup/gettext_setup.rb#L29) to use PO files, and therefore will not correctly parse the proposed subdirectories. I am happy to refactor the gem if that seems reasonable to achieve a common directory structure between projects.

file_paths.md
MikaelSmith: The only prior example I see is in the Ruby install, where it includes a `share` directory. So `share/locale` would be consistent with that and the Linux paths.

file_paths.md
```diff
@@ -298,6 +298,10 @@ create a `puppet` user or group.
         lib *
             puppet.rb *
         locale *
+            <lang_COUNTRY>
```
glennsarti: Should we have an example of what it would look like for say, en-US & puppet?

Magisus: Added some, is that what you were thinking?

glennsarti: Yes it was....thanks

MikaelSmith:
I'm not tied to .po vs .mo, I think we could say both and leave it up to each project which they use.

Overriding the C++ tooling (which requires `<lang_COUNTRY>/LC_MESSAGES/<project>.mo`) would be hard.

Magisus:
I am personally in favor of consistency across projects. Researching how these files are stored in general was very difficult because there seems to be no agreed-upon way to do it. I'd rather we didn't replicate that problem internally, at least not within the puppet-agent package. So unless there is a good, specific reason to use PO files over MO files for Ruby, I think we should lay down the directory structure for MO and then use it everywhere. And because that format is required for the C++ projects, we will have to write out automation for compiling the PO files anyway.

file_paths.md
Magisus: After poking around on Windows for a while, it turns out we already have a `share` directory alongside `bin` and `lib` in `C:\Program Files\Puppet Labs\Puppet\puppet`, which for some reason isn't listed here. I agree with Michael, it makes sense to put `locale` in there for consistency.

Iristyle:
In agreement that `.mo` looks to be the way to go.

file_paths.md
```diff
@@ -292,6 +300,14 @@ create a `puppet` user or group.
             puppet *                          # ruby bin wrapper
         lib *
             puppet.rb *
+        share *
+            locale
```
Iristyle: Anything created at installation time should have a `*` next to it.

A lack of a `*` generally indicates Puppet creates (and typically manages) the file / directory on first run.  I believe this entire path structure is created by installing a package (for both Windows and Linux)

Magisus: I wasn't sure if it needed to be on every level, updated!

geoffnichols:
👍

Magisus:
@Iristyle did you have any further comments on this? I have verified that these paths can be correctly installed to and read from, on both Linux and Windows.

Iristyle:
Sorry, I was good with this as well. 👍

PR #93, opened 2017-03-08
Resource API v1.0

DavidS:
Public PR to facilitate conversations. 

This contains draft spec language, code examples, and the history how we got here.

There is still much to do before this can be merged.


lutter:
Could this be merged with a big 'Draft - under discussion' disclaimer ? (If need be on a separate branch) That way, others can make PR's against the draft to amend it.

language/resource-api/apt_key.rb
```diff
@@ -0,0 +1,375 @@
...
+            type:    'Pattern[/\A((hkp|http|https):\/\/)?([a-z\d])([a-z\d-]{0,61}\.)+[a-z\d]+(:\d{2,5})?$/]',
+            docs:    'The key server to fetch the key from based on the ID. It can either be a domain name or url.',
+            default: 'keyserver.ubuntu.com'
+        },
+        options:     {
+            type: 'Optional[String]',
+            docs: 'Additional options to pass to apt-key\'s --keyserver-options.',
+        },
+        fingerprint: {
+            type:      'String',
```
igalic: shouldn't this then be `String[40]`?

language/resource-api/apt_key.rb
```diff
@@ -0,0 +1,375 @@
...
+            docs:      'The 8-digit hexadecimal id of the specified GPG key.',
+            read_only: true,
+        },
+        expired:     {
+            type:      'Boolean',
+            docs:      'Indicates if the key has expired.',
+            read_only: true,
+        },
+        expiry:      {
+            # TODO: should be DateTime
```
igalic: :+1:

language/resource-api/apt_key.rb
```diff
@@ -0,0 +1,375 @@
...
+            docs:      'The date the key will expire, or nil if it has no expiry date, in ISO format.',
+            read_only: true,
+        },
+        size:        {
+            type:      'Integer',
+            docs:      'The key size, usually a multiple of 1024.',
+            read_only: true,
+        },
+        type:        {
+            type:      'String',
```
igalic: `Enum[rsa, dsa, ecc, ecdsa]`

language/resource-api/apt_key.rb
```diff
@@ -0,0 +1,375 @@
...
+    def prune_parameters(*args)
+      puts "not pruning #{args.inspect}" if args.length > 0
+      self
+    end
+
+    def to_manifest
+      [
+          "apt_key { #{values[:name].inspect}: ",
+      ] + values.keys.select { |k| k != :name }.collect { |k| "  #{k} => #{values[k].inspect}," } + ['}']
+    end
```
igalic: this feels very clear and logical, but the rest up there just reads like boilerplate…
what am i missing?

DavidS:
@lutter people can always PR against my branch/fork; collaborators should even be able to push to my branch directly

@igalic thanks for your comments! I'll work them in on my next revision - maybe later this week.

language/resource-api/apt_key.rb
DavidS: Yup. I'll change it to `Pattern[/[a-f0-9]{40}/]`, as this read-only attribute will always have a specific form. Same for the next two.

language/resource-api/apt_key.rb
DavidS: The whole file is a mixture of example and prototype library code, I've added an explanation at the top:

```
# This is a experimental hardcoded implementation of what will be come the Resource API's runtime
# environment. This code is used as test-bed to see that the proposal is technically feasible.
```

the `simpleapt.rb` is closer to what a real provider would look like.

DavidS:
To keep the loop closed, I'm currently working on a prototype implementation of this in https://github.com/DavidS/puppetlabs-apt/tree/resource-api-experiments

It currently only supports listing `apt_key` resources through `puppet resource`. More to come.

reidmv:
Regarding e974464: how we handle `noop` is critically important to sustaining an important benefit of Puppet, which is that the framework handles making sure when you run in noop no changes are made to the system. This is a key differentiator between our noop feature and e.g. Chef's dry-run, because more onus is put on Chef's resource developers to make sure dry-run doesn't change the system.

I haven't looked at this recent change it detail so maybe this is already true, but I believe it's important the _default, simplest behavior_ should be noop-compatible. That if you want to do anything that would not be noop-safe, that should require more typing or parameter passing as an opt-in "I know what I'm doing" to bypass noop safeguards. That is, command invocation should be noop-safe by default, and you should have to ask to step outside those safeguards.

Is that the case with the changes in e974464?

DavidS:
@reidmv : by default, the `set` function will only be called if the Runtime Environment wants the Implementation to affect changes in the target system. This has not changed from the original API. What has changed, is that Implementations can opt into taking over that decision responsibility themselves, by declaring the `noop_handler` feature. In this case the Commands API provides a simple `..., noop: noop)` parameter (that can easily be checked [even mechanically]) to defang commands that should not run under noop, while still leaving the possibility to run commands when necessary. To note: many providers in the future are expected to talk to APIs (not local commands) which will have no need for `noop_handler` at all, or require different safety measures.

@lutter : this seems like only a reshuffling of symbols.

language/resource-api/README.md
```diff
@@ -217,79 +217,118 @@ You can set default environment variables on the `@cmd.environment` Hash, and a
...
-The commands take a `noop:` keyword argument, and will signal success while skipping the real execution if necessary.
+By default the `stdout` of the command is logged to debug, while the `stderr` is logged to warning.
+
+#### Implementing `noop` for `noop_handler`
+
+The `run` method takes a `noop:` keyword argument, and will signal success while skipping the real execution if necessary. Providers implementing the `noop_handler` feature should use this for all commands that are executed in the regular flow of the implementation.
+
+```ruby
+class Puppet::Provider::AptKey::AptKey
+  def set(context, changes, noo: false)
```
kbarber: noo

DavidS:
While it's not yet fully implemented, I'd like to call for a final round of comments on the text as it stands.

A 0.9 of the core concepts is available as https://rubygems.org/gems/puppet-resource_api and is already being trialled by us internally.

DavidS:
PS: see also `pdk new provider` in the upcoming v1.4 and the pre-releases available on http://nightlies.puppet.com/

DavidS:
The Resource API v1.0 gem has been released, so this is now a Thing[tm].

PR #97, opened 2017-08-03
(PCP-774) Add tasks-cache to puppet-specifications

ekinanp:


file_paths.md
```diff
@@ -146,6 +146,7 @@ The files annotated by an '*' indicate that they are created by package installa
         modules *
             pxp-module-puppet *
         spool *                            # directory containing results of pxp-agent modules
+        tasks-cache *                      # directory containing a cache of all the files required for a specific task
```
MikaelSmith: Not accurate. It contains a cache of files for all downloaded tasks.

file_paths.md
```diff
@@ -252,6 +253,7 @@ create a `puppet` user or group.
                 pxp-agent.log                 # enabled by default
                 pxp-access.log                # not enabled by default
             spool *                           # directory containing results of pxp-agent modules
+            tasks-cache *                     # directory containing a cache of files for all downloaded tasks
```
ekinanp: @MikaelSmith Is this what you meant?

MikaelSmith: Yes. On Windows, the tasks-cache wasn't in the var directory, it was at pxp-agent/tasks-cache.

file_paths.md
```diff
@@ -253,6 +254,7 @@ create a `puppet` user or group.
                 pxp-access.log                # not enabled by default
             spool *                           # directory containing results of pxp-agent modules
             run *
+        tasks-cache *                         # directory containing a cache of files for all downloaded tasks
```
ekinanp: @MikaelSmith

Iristyle: I'm wondering if this should be a sibling directory to `spool` under `var`, so that's closer to the *nix layout?

Iristyle: I realize @MikaelSmith asked you to change this in a prior comment... but I'm wondering what went into that decision?

MikaelSmith: Putting the `spool` dir under `var` on Windows looked like a mistake to me. On *nix it'd be under `opt`. However, we could choose to make it consistent even though it's kind of weird.

MikaelSmith: Should I propose moving the spool dir as well? Any objection to moving that direction, anyway?

PR #98, opened 2017-09-20
(PUP-7965) Update language specification for "hash without braces"

thallgren:


language/expressions.md
```diff
@@ -227,6 +227,20 @@ A "Literal" Hash has the following syntax.
...
+are converted into one single `Hash` as if they were surrounded by curly braces.
+
+Examples:
+
+     $a = [1, 2, a => 3, b => 4, 5]              # $a becomes [1, 2, {a => 3, b => 4}, 5]
+     foo('hello', a => 'universe', b => 'world') # same as   foo('hello', { a => 'universe', b => 'world' })
+     Struct[a => Integer, b => String]           # same as Struct[{a => Integer, b => String}]
+
+It is illegal to use `HashEntry` expressions in the argument list of a top level function call unless the list
+is delimited with parenthesis.
```
hlindberg: I don't think we refer to "function call without parentheses" as "top level function call" as they are not required to be in something you can think of as "top level". Maybe say "HashEntry expressions can only be  used in a function call argument list when the arguments are enclosed in parenthes**es**."  (parenthes**is** is in singluar).

thallgren: I used you suggested wording with a slight alteration to get rid of the ambiguity whether it's the list or each individual argument that's being enclosed in parentheses.

PR #99, opened 2017-09-26
(PCP-792) Add non-root paths for pxp-agent

MikaelSmith:


file_paths.md
```diff
@@ -368,6 +368,15 @@ On Windows, when not running on the SYSTEM account
 
 These sections describe other Puppet packages that rely on puppet-agent to create the initial directory layout. It does not attempt to specify the full set of file paths for these packages, just cases where the other package has a dependency on puppet-agent.
 
+pxp-agent also supports running as non-root, and uses the following paths.
+
+    ~/.puppetlabs/etc/pxp-agent/{pxp-agent.conf,modules}
```
mruzicka: The `{,}` syntax is not used anywhere else in the file, mind discarding it at the expense of making the section more verbose?

MikaelSmith: I wanted something more descriptive than the section above, which has named settings for each dir. I'll see how it looks following the patterns higher up.

MikaelSmith: Updated, does that read better?

PR #100, opened 2017-10-30
(ORCH-2119) Add task specification

MikaelSmith:
Adds the version 1 task specification and related schema.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -1,0 +1,219 @@
+# Puppet Task Spec (Version 1)
+
+The Puppet Task ecosystem allows users to execute actions on target systems. A **Task** is the smallest component of this system capturing a single action that can be taken on a target and is - in its simplest form - an executable file. It is packaged and distributed as Puppet modules which are made available to the task runner.
+
+## Design goals
+
+- The barrier to entry for writing simple tasks should be as low as possible. Drop a script in a folder and it's runnable.
+- Start with a small functional core and add features as needed.
+
+## Terminology
+
+**Task**: A task is a file and an optional metadata file that will be executed on a target.
+
+**task runner**: The application which executes tasks on target systems. This is the interface for controlling and running tasks, and is assumed to provide some sort of programatic interface (API). In initial releases this is Orchestrator (executing via PXP) and [Bolt] (via ssh/winrm). In the future new task runners may be added. For example the puppet agent may run tasks from resources.
+
+**UI**: User interface, such as CLI tools ([Bolt], `puppet task`) or the Puppet Enterprise Console.
+
+## Puppet Task Spec Versioning
+
+Orchestrator shipped in PE 2017.3 supports version 1. [Bolt] is planned to support version 1 by 1.0.
+
+Changes to the task spec that do not bump the version must not break existing tasks. We may add properties to the metadata defined in version 1 without bumping the version.
+
+Older versions of the task runner may not support tasks that rely on newer features of the task spec. The task spec version is not intended to capture this now. Authors should use the `puppet_task_version` field in the module's metadata to document such incompatibilities for users.
+
+## Task packaging and file
+
+Tasks are packaged and distributed in the `/tasks` directory of a Puppet module. Any file in the top level of that directory with a valid name is considered a task. In addition to tasks, task authors may create a metadata file for the task with the same name and a `.json` extension.
+
+### Task name and filename
+
+- Task names have the same restriction as puppet type names and must match the regular expression `\A[a-z][a-z0-9\_]*\z`.
+- Filenames not matching the task name regular expression will be ignored.
+- Tasks are referred to as `module_name::task_name`.
+- Tasks must exist at the top level of the tasks directory to be found.
+- Task metadata is stored in `task_name.json` and is optional.
+- The executable for the task must be stored in `task_name` or `task_name.<extension>`(where extension != `json`, `md` or `conf`).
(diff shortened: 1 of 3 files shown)


tasks/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+The task file is copied to the target and then executed on the target by the task runner.
+
+No arguments are passed to the task when it is executed.
+
+The task file in the module does not need execute permissions set.
+
+The location of the task file varies based on the transport used by the task runner. Task authors have no control over this path.
+
+The environment the task is executed in is also controlled by the task runner. Task authors should document any requirements they have.
```
MikaelSmith: @adreyer `environment` is an overloaded term, do you have suggestions on how to be more specific? Is this system environment variables or general operating environment (as in what's installed, what you have access to)? Doesn't seem like Puppet environment.

adreyer: I think it's primarily the system environment variables. like what's in $PATH etc.

tasks/README.md
```diff
@@ -0,0 +1,244 @@
...
+{
+  "a": 1,
+  "b": "a string",
+  "c": [1, 2, "3"],
+  "d": {"x": {"y": [0]}}
+}
+```
+
+##### Environment Variables
+```
```
adreyer: bash?

tasks/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+The `powershell` input method will pass each param as a named argument to the powershell script. This is the default input method for task files ending with a `.ps1` extension.
+
+## Output handling
+
+### Stdout
+
+Task output is consumed on stdout and made available through the task runner API.
+
+If the output cannot be parsed as a JSON object, one is created by the task runner with the captured stdout stored in the  `_output` key: i.e. `{"_output": "the string of output"}`. Otherwise the parsed JSON object is used.
```
adreyer: We should clarify that JSON object means object not array or string.

tasks/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+Reserved output fields:
```
adreyer: All _ prefixed keys are reserved and should only be used as described below. Should we just say top level keys must match the param regex or be listed below?

tasks/README.md
```diff
@@ -0,0 +1,244 @@
...
+- **_output**: this will contain the stdout of the task if it returns plaintext output
```
adreyer: A text summary of the job result. If the job does not return a JSON object the contents of stdout will be put in this key.

tasks/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+Errors should be generated in the task_runner when the task could not be successfully executed or the results of the task could not be captured. 
+
+Some examples:
+- **task_file_error**: the task could not be copied onto the target
+- **unexecutable_task**: The task execution failed 
+- **unparsable_output**: The task output could not be parsed with any of the specified formatters
+- **output_encoding_error**: The task output was not utf-8
+
+The JSON schema for errors:
```
adreyer: I think we should put this in another file.

tasks/README.md
```diff
@@ -0,0 +1,244 @@
...
+```
+{
+  "$schema": "http://json-schema.org/schema#",  
+  "description": "The format for errors in puppet",
+  "type": "object",
+  "properties" : {
+    "type": {
```
adreyer: This is kind right?

tasks/task.json
```diff
@@ -0,0 +1,61 @@
...
+    "input_method": {
+      "type": "string",
+      "enum": ["stdin", "environment"],
+      "description": "What input method should be used to pass params to the task"
+    },
+    "parameters": {
+      "$ref": "#/definitions/paramsObject",
+      "description": "An object mapping valid parameter names to corresponding json-schemas"
+    },
+    "output": {
```
adreyer: lets cut this for now

tasks/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+**puppet_task_version**: The version of this spec used.
+
+**supports_noop**: Default `false`. This task respects the "\_noop" metaparam. If this is not set the task runner will refuse to run the task in noop.
+
+**input_method**: What input method to use to pass parameters to the task. Default varies, see [Input Methods](#input-methods).
+
+**parameters**:  The parameters or input the task accepts listed with a [Puppet data type](../language/types_values_variables.md) string and optional description. Top level params names have the same restrictions as Puppet class param names and must match `\A[a-z][a-z0-9_]*\z`.
+
+**output**: A description of the fields in the tasks output if it returns structured data. Names for top level output keys must match the regex `\A[a-z][a-z0-9_]*\z`. This exists not to validate output but to generate documentation and provide formatting hints to the UI. Keys beginning with '_' are reserved for internal metaparams (for instance `_output`).
```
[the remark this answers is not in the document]
MikaelSmith: The task schema we've provided includes the `output` field. Some comment seems useful, but it could be simpler and say "may be used in the future".

tasks/README.md
MikaelSmith: We don't have anything to enforce that.

tasks/README.md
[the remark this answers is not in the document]
MikaelSmith: Ah, you mentioned removing it from the schema. I'll do that.

MikaelSmith:
Updated in a new commit. I'd like to squash them before (or on) merge.

MikaelSmith:
Squashed.

PR #106, opened 2018-03-14
(PUP-4380) Describe plan for moving non-core modules out of puppet

joshcooper:
Describes the user stories, file paths, and technical issues around extracting non-core types and providers from puppet.

### DIFF
--- a/moving_modules.md
+++ b/moving_modules.md
@@ -1,0 +1,224 @@
+Moving Non-Core Types & Providers
+=================
+
+## Index
+
+* [Goals](#goals)
+* [User Stories](#user-stories)
+* [File Path Changes](#file-path-changes)
+* [Resource Type Taxonomy](#resource-type-taxonomy)
+  * [Internal](#internal)
+  * [Core](#core)
+  * [External](#external)
+  * [Dependencies](#dependencies)
+* [Packaging](#packaging)
+
+# Goals
+
+* Remove non-core types and providers from the puppet repo and move them to modules. Doing so will reduce the surface area of puppet, decrease puppet CI cycle times, while making the extracted types and providers more accessible to community members and increasing their maintainability.
+* Continue to bundle some modules with puppet-agent during packaging so that users have a batteries-included experience.
+
+# User Stories
+
+## Serverless Puppet
+
+As an admin, I want to run masterless puppet and immediately be able to manage basic resources on my system, so that I can more quickly get value from puppet. I do not want to search the forge in order to perform basic system tasks. The preinstalled modules should make sense for the local platform I'm running puppet on. On *nix, I should be able to manage cron, and on Windows, powershell resources.
+
+## Server-based Catalog Compilation
+
+As an admin, I should be able to compile catalogs for resources whose types are included in the locally installed puppet-agent package, so that I don't have to install modules I already have.
+
+## Server-based Catalog Application
+
+As an admin, if catalog compilation succeeds, then I want assurance that the catalog will be applied consistently across all agent versions regardless of which preinstalled modules are on each agent. For example, if the master has a newer version of a type, then all agents in that environment should use the same version of the type (and its provider) at catalog application time.
+
+## Module Updates
+
+As an admin, if there is a problem with a type/provider packaged with puppet-agent, I want to be able to install a newer version from the forge (for both serverless and server-based), so that I don't need to wait for Puppet to release a new puppet-agent build.
+
+ ## Puppet-Agent Updates
+
+As an admin, when the puppet-agent package is updated, I want to be able to use the new versions of types and providers that come with the new package, even if that means overwriting and deleting older preinstalled versions. However, installing a new puppet-agent package should not clobber modules I've installed via puppet module tool, r10k, etc.
+
+## Module Pinning
+
+As an admin, if I update the puppet-agent package, and it updates a preinstalled module, but the module introduces a regression, I want to easily install the older version of the module from the forge.  I don't want to be forced to downgrade puppet-agent packages, because that process introduces more risk.
+
+# File Path Changes
+
+Puppet's default `basemodulepath` includes two module directories visible to all environments:
+
+    /etc/puppetlabs/code/modules
+    /opt/puppetlabs/puppet/modules
+
+The first is where modules are typically installed to via puppet module tool, r10k, codemanager/filesync. The second path typically contains modules that are installed in PE environments, though there is nothing stopping users from manually installing modules there, e.g. `puppet module tool install puppetlabs-apt --target-dir /opt/puppetlabs/puppet/modules`. It is important that preinstalled modules do not use those same locations, otherwise, it will confuse package managers.
+
+We propose a new directory `/opt/puppetlabs/puppet/vendored_modules` to be created at puppet-agent installation time and containing all modules added to puppet-agent at packaging. The directory should be appended to the default `basemodulepath` so that the modules are available during compilation and application.
+
+# Resource Type Taxonomy
+
+## Internal
+
+The following types are internal to Puppet and will remain as is:
+
+    component
+    schedule
+    stage
+    whit
+
+## Core
+
+The following types will be left in Puppet for now. The `file`, `user`, and `group` types are needed to apply settings catalogs. The `filebucket`, `resources`, and `tidy` types know too much about puppet internals to be removed. The `notify` type is used extensively in puppet rspec tests as it is the most basic (providerless) type. The `package` and `service` types have multiple providers for each type, which makes removal more difficult. We may remove them at a later time (TBD).
+
+    exec
+    file
+    filebucket
+    group
+    notify
+    package
+    resources
+    service
+    tidy
+    user
+
+## External
+
+The following types and providers will be removed from Puppet. A subset (details TBD below) will be added back to puppet-agent during packaging.
+
+Each top-level path below specifies the name of the module, e.g. `augeas`, and the files contained within each module. The modules will be installed in a new directory visible to Puppet's autoloader, so catalog compilation and application will just work without additional configuration. Puppet will prefer modules in the modulepath and pluginsync'ed lib directory over the packaged modules, so that newer versions of modules can fix bugs in packaged modules.
+
+    Path                                           Comments
+
+    <vendored_modules>/                            (*nix) /opt/puppetlabs/puppet/vendored_modules/
+                                                   (Windows) C:\Program Files\Puppet Labs\Puppet\puppet\vendored_modules
+
+      augeas/                                      Depends on 'puppet/parameter/boolean'
+        lib/puppet/feature/augeas.rb               Extracted from lib/puppet/features/base.rb        
+        lib/puppet/type/augeas.rb
+        lib/puppet/provider/augeas/augeas.rb
+
+      cron/                                        Depends on 'puppet/provider/parsedfile',
+        lib/puppet/type/cron.rb                               'puppet/util/filetype'
+
+      host/                                        Depends on 'puppet/property/ordered_list',
+        lib/puppet/type/host.rb                               'puppet/provider/parsedfile',
+        lib/puppet/provider/host/parsed.rb
+
+      k5login/                                     Depends on 'puppet/type/file/selcontext',
+        lib/puppet/type/k5login.rb                            'puppet/util/selinux'
+                                                              
+      mailalias/                                   Depends on 'puppet/provider/parsedfile'
+        lib/puppet/type/mailalias.rb
+        lib/puppet/provider/mailalias/aliases.rb
+     
+      maillist/
+        lib/puppet/type/maillist.rb
+        lib/puppet/provider/maillist/maillist.rb
+
+      macdslocal/                                  Depends on 'puppet/provider/nameservice/directory_service',
+        lib/puppet/type/                                      'puppet/util/plist'
+          computer.rb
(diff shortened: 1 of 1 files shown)


moving_modules.md
```diff
@@ -0,0 +1,224 @@
...
+  * [External](#external)
+  * [Dependencies](#dependencies)
+* [Packaging](#packaging)
+
+# Goals
+
+* Remove non-core types and providers from the puppet repo and move them to modules. Doing so will reduce the surface area of puppet, decrease puppet CI cycle times, while making the extracted types and providers more accessible to community members and increasing their maintainability.
+* Continue to bundle some modules with puppet-agent during packaging so that users have a batteries-included experience.
+
+# User Stories
```
ahpook: these stories are good but they are all about end user - should we include some around the maintainership of modules post-extraction? something like 

* as an experienced puppet user, when i fix a bug in a puppet-maintained provider i use, i want to contribute that fix back upstream with a minimum of friction, so that i don't have to carry the patch to future versions of the agent

* as a puppet community member with expertise in parts of the provider ecosystem, i want to manage the flow of fixes and contributions into the providers i care about, so that contributions and module releases can happen rapidly without requiring puppet core contributors to do a full agent release.

... or so?  those might be outside scope but it'd be good to see them incorporated *somewhere* if not here.

moving_modules.md
```diff
@@ -0,0 +1,224 @@
...
+# File Path Changes
+
+Puppet's default `basemodulepath` includes two module directories visible to all environments:
+
+    /etc/puppetlabs/code/modules
+    /opt/puppetlabs/puppet/modules
+
+The first is where modules are typically installed to via puppet module tool, r10k, codemanager/filesync. The second path typically contains modules that are installed in PE environments, though there is nothing stopping users from manually installing modules there, e.g. `puppet module tool install puppetlabs-apt --target-dir /opt/puppetlabs/puppet/modules`. It is important that preinstalled modules do not use those same locations, otherwise, it will confuse package managers.
+
+We propose a new directory `/opt/puppetlabs/puppet/vendored_modules` to be created at puppet-agent installation time and containing all modules added to puppet-agent at packaging. The directory should be appended to the default `basemodulepath` so that the modules are available during compilation and application.
```
ahpook: how about `vendor_modules` for parallelism with ruby's `vendor_ruby` area?

moving_modules.md
```diff
@@ -0,0 +1,224 @@
...
+The following types are internal to Puppet and will remain as is:
+
+    component
+    schedule
+    stage
+    whit
+
+## Core
+
+The following types will be left in Puppet for now. The `file`, `user`, and `group` types are needed to apply settings catalogs. The `filebucket`, `resources`, and `tidy` types know too much about puppet internals to be removed. The `notify` type is used extensively in puppet rspec tests as it is the most basic (providerless) type. The `package` and `service` types have multiple providers for each type, which makes removal more difficult. We may remove them at a later time (TBD).
```
ahpook: i suggest using the same language as the intro - "extract" instead of "remove", so it doesn't feel like they are being *deleted*, just pulled into a different place.

moving_modules.md
```diff
@@ -0,0 +1,224 @@
...
+    notify
+    package
+    resources
+    service
+    tidy
+    user
+
+## External
+
+The following types and providers will be removed from Puppet. A subset (details TBD below) will be added back to puppet-agent during packaging.
```
ahpook: s/removed/extracted/

moving_modules.md
```diff
@@ -0,0 +1,224 @@
...
+# Goals
+
+* Remove non-core types and providers from the puppet repo and move them to modules. Doing so will reduce the surface area of puppet, decrease puppet CI cycle times, while making the extracted types and providers more accessible to community members and increasing their maintainability.
+* Continue to bundle some modules with puppet-agent during packaging so that users have a batteries-included experience.
+
+# User Stories
+
+## Serverless Puppet
+
+As an admin, I want to run masterless puppet and immediately be able to manage basic resources on my system, so that I can more quickly get value from puppet. I do not want to search the forge in order to perform basic system tasks. The preinstalled modules should make sense for the local platform I'm running puppet on. On *nix, I should be able to manage cron, and on Windows, powershell resources.
```
ahpook: don't know if it matters in the real renderer, but this unmatched, unescaped `*` causes the whole file to be italicised in github's markdown display :(

moving_modules.md
```diff
@@ -0,0 +1,224 @@
...
+                                                              
+      mailalias/                                   Depends on 'puppet/provider/parsedfile'
+        lib/puppet/type/mailalias.rb
+        lib/puppet/provider/mailalias/aliases.rb
+     
+      maillist/
+        lib/puppet/type/maillist.rb
+        lib/puppet/provider/maillist/maillist.rb
+
+      macdslocal/                                  Depends on 'puppet/provider/nameservice/directory_service',
```
ahpook: i think `provider/nameservice/directory_service` should come along for the ride here

joshcooper: Unfortunately that provider is the parent for user and group on osx also, and moving those out is complicated by need to manage settings catalogs and that the resources have multiple providers...

moving_modules.md
joshcooper: Good point, thoughts @branan?

branan: neutral response

moving_modules.md
branan: Probably depends on the markdown engine used. We should escape to be safe

moving_modules.md
joshcooper: I'm still vague on some of the packaging details, and the community part is even farther removed, but I agree having the stories helps frame the problem space better.

joshcooper: Also puppet as a gem and puppet type generate

hlindberg:
A thought - loading things at runtime requires quite a lot of IO - if we could generate an index at "some time" that is not runtime we can load that instead of scanning the file system. Vendored modules included in a package could have this as it is not expected that users hack on those when they are on the file system. We could perhaps use the same when installing via module-tool. Reason for me posting this is that puppet prepopulates loaders with known resource types  - when they move out to a module, they are not really known and needs to be loaded the same way as everything else. We also do not know if different versions of vendored modules are used in different environments since you can override the vendored - that opens up for problems with environment isolation - and it would be beneficial to also include pre-generated "generate types" to make the types env-safe.

From a use case perspective - "I don't want to worry about environment isolation when I need to override a vendored module", "I don't want to see performance degradation in my compiles"

joshcooper:
Good points @hlindberg, yes, removing T&P into modules combined with installing different versions of those modules in different environments would make env isolation worse for anyone not using `puppet generate types` workflow. I will update the use case section.

I agree precomputing that information would be good. I think that would be best handled in `pdk build` (which will supersede PMT) or in r10k/codemanager. Now that you can HUP opensource puppetserver when new code has been deployed (and PE already has that capability via HUP or codemanager), it should be easier to move in an "everything is pregenerated" mode of operation.

hlindberg:
For this to work well, we need to also make use of pregenerated things per module. The `generate types` currently handles the entire environment and stores the information in a central location. Irrespective of which tool it is building the precomputed information, it needs to be using indexing that is appropriate for a particular version of puppet as I don't think we will stop innovating in the "things we need to load" domain. When we do have these indexes, it is will mean a lot performance wise as as we learned recently - a scan per module to get candidate files is quite expensive.

moving_modules.md
joshcooper: Is the name of the module ok? Wasn't sure if it should be `dslocal` or `osxdslocal`, etc..?

joshcooper:
@ahpook @hlindberg I added some revisions in a second commit to make review easier.

joshcooper:
There are already forge modules with the same module names as what I'm proposing:

https://forge.puppet.com/camptocamp/augeas
https://forge.puppet.com/ericsson/cron
https://forge.puppet.com/lgbarn/mount
https://forge.puppet.com/puppet/selinux

Although the modules are namespaced on the forge, you can't install 2 modules of the same module name from different authors, e.g. if you've already installed `camptocamp-augeas` then trying to install another augeas module fails:

```
$ bx puppet module install svarrette-augeas
Notice: Preparing to install into /Users/josh/.puppetlabs/etc/code/environments/production/modules ...
Notice: Downloading from https://forgeapi.puppet.com ...
Error: Could not install module 'svarrette-augeas' (latest)
  Installation would overwrite /Users/josh/.puppetlabs/etc/code/environments/production/modules/augeas
    Currently, 'camptocamp-augeas' (v1.6.0) is installed to that directory
    Use `puppet module install --force` to install this module anyway
```

So I think we will need to create unique names for all extracted modules, maybe something like `puppet-augeas_core`, `puppet-cron_core`, `puppet-mount_core`?

mciurcio:
I've being using Puppet since 0.25 and I do training, consulting, deployment (PE and OSS) for a living since then.

I think this proposal is adding extra complexity for little to no actual again.

Puppet never had a "batteries included" experience since forever, IMHO. It is unfeasible to do anything efficiently without at last close to 20 modules from the Forge from the get go:
- stdlib,
- archive
- epel
- staging
- sudo
- concat
- ntp
- powershell
- apt
- inifile
- java_ks
- java
- augeasproviders_*
- registry
- acl
- chocolatey

So Puppet IMHO is already pretty much close to bare bones compared to other solutions, no batteries included at all. A couple of more modules to install as a base to start actually doing something is completely reasonable, in exchange for not adding all this complexity.

Honestly all User Stories written on this PR are a nightmare from a workflow and maintenance perspective and I would not hesitate for a second to tell people/students/clients to do not do it.

Also, aren't we supposed to stop mixing Agent and Server stuff? I see only more scenarios here were this could explode on my face, as has happened in the past.

joshcooper:
@mciurcio what are you proposing as an alternative? Don't vendor any modules in puppet-agent, and leave it up to users to install modules they care about?

mciurcio:
@joshcooper Yep. Don't vendor, just publish the modules on the Forge. Sorry if my previous comment sounded harsh, it is not my intention.

trevor-vaughan:
The only way that I can see this working well is if Puppet could incorporate the author name as part of the module namespace automatically and allow modules to co-exist with the same base names. It probably is far too much complexity for daily maintenance. I wouldn't mind loading speed gains if those are a possibility though.

trevor-vaughan:
@joshcooper I'd like to add one more: As an experienced Puppet user, I would like to be able to use the `puppet-build-tools` to build the component releases so that I can test the 'as expected' changes.

moving_modules.md
```diff
@@ -0,0 +1,317 @@
...
+
+      zfs_core/
+        lib/puppet/type/zfs.rb
+        lib/puppet/provider/zfs/zfs.rb
+
+      zone_core/                                   Depends on 'puppet/property/list'
+        lib/puppet/type/zone.rb
+        lib/puppet/provider/zone/zone.rb
+
+      zpool_core/
```
jhelwig: This is being rolled into the zfs module.

joshcooper: updated

moving_modules.md
```diff
@@ -0,0 +1,315 @@
...
+issues. Users that install newer versions of modules containing types that are
+also vendored, should use `puppet generate types` to ensure types in one
+environment don't affect other environments. This isn't really a new concern,
+it's just that users haven't had to `puppet generate types` for builtin types
+before.
+
+## Module Dependencies
+
+The autoloader does not rely on module metadata to load types. So any module
+that relies on an type, should just work in Puppet 6 provided the extracted
```
rhowe-gds: > relies on a type

moving_modules.md
```diff
@@ -0,0 +1,315 @@
...
+
+Agents may have different versions of vendored modules than the server used to
+compile the catalog. The agent should always use the pluginsync'ed version of
+the type and providers instead of whatever version is present in the agent's
+puppet-agent package.
+
+## Agent Compatibility
+
+Pre-6.0 agents will always prefer the builtin version over what was
+pluginsynced. If a new property/parameter is added to the type on the master and
```
jhelwig: Unfortunately, pre-6.0 agents currently prefer the pluginsync'ed version over the built-in one.

joshcooper:
To watchers, the latest commit reflects our current thinking. Vendored modules will effectively be part of the ruby load path like they are today, and they will not be pluginsynced. See https://github.com/puppetlabs/puppet/pull/6917

New users will get batteries included. Advanced users wanting to manage all module versions can set `vendormoduledir = ''` and puppet will ignore its vendored modules.

moving_modules.md
```diff
@@ -0,0 +1,358 @@
...
+
+## External
+
+The following types and providers will be extracted from Puppet. A subset (details TBD below) will be added back to puppet-agent during packaging.
+
+Each top-level path below specifies the name of the module, e.g. `augeas`, and the files contained within each module. The modules will be installed in a new directory visible to Puppet's autoloader, so catalog compilation and application will just work without additional configuration. Puppet will prefer modules in the modulepath and pluginsync'ed lib directory over the packaged modules, so that newer versions of modules can fix bugs in packaged modules.
+
+    Path                                           Comments
+
+    <vendor_modules>/                              (*nix) /opt/puppetlabs/puppet/vendor_modules/
```
shrug: Should this also get added to https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md ?

joshcooper: Good point, I will update that.

PR #108, opened 2018-03-23
(BOLT-394) Document spec for cross-platform tasks

nicklewis:


nicklewis:
This is under debate...

tasks/README.md
```diff
@@ -33,12 +33,11 @@ Tasks are packaged and distributed in the `/tasks` directory of a Puppet module.
 - Filenames not matching the task name regular expression will be ignored.
 - Tasks are referred to as `module_name::task_name`.
 - Tasks must exist at the top level of the tasks directory to be found.
-- Task metadata is stored in `task_name.json` and is optional.
-- The executable for the task must be stored in `task_name` or `task_name.<extension>`(where extension != `json`, `md` or `conf`).
+- Task metadata is stored in `task_name.json`.
```
DavidS: Neither the ticket nor the commit messages contain a rationale for making task metadata mandatory, and it does not seem to be required for the rest of the additions here.

DavidS: :+1:

tasks/README.md
```diff
@@ -98,9 +103,17 @@ Any parameter that does not specify a type will accept any value (default type i
...
 If a parameter type accepts `null` the task runner will accept either a `null` value or the absence of the property. Task authors must accept missing properties for nullable parameters. Task authors must not differentiate between absent properties and properties with `null` values.
 
+### Metaparameters
+
+In addition to the tasks parameters, the task runner may inject metaparameters prefixed by '_'. These include `_noop` and `_task`.
+
 ## Task execution
 
-The task file is copied to the target and then executed on the target by the task runner.
+If the task has multiple executable files, the `implementations` field of the metadata is used to determine which executable is suitable for the target. Each implementation can specify `requirements`, which is an array of the required "features" to use that implementation. The available features are defined by the task runner.
```
adreyer: We should document the features that task runners should provide.

We should document that task authors should name their tasks to supports backwards compatibility.

nicklewis: Do you think that latter point belongs in the spec or in the writing tasks doc?

nicklewis:
Updated to hopefully clarify the explanation of how task metadata/implementations are associated, and also to call out the features that task runners should provide.

DavidS:
That reads so much nicer now!

tasks/README.md
```diff
@@ -29,22 +29,19 @@ Tasks are packaged and distributed in the `/tasks` directory of a Puppet module.
...
+A task consists of an optional metadata file and one or more implementation files. Files with the `.json` extensions are metadata files, and any other file extension (including files with no extension) is an implementation file. Implementation files do not need the executable bit set.
 
-A task with multiple executable files will be considered invalid.
+An implementation file `<task>.<ext>` without a corresponding metadata file `<task>.json` is a task.
 
-Tools for validating this metadata and making local task execution easier when testing will eventually be added to the PDK.
+A metadata file `<task>.json` also identifies a task. If the metadata specifies an `implementations` array, then the files listed in that array are the implementations for the task. Otherwise, there must be one corresponding implementation file `<task>.<ext>`.
+
+Task names have the same restriction as puppet type names and must match the regular expression `\A[a-z][a-z0-9\_]*\z`. The extensions `.md` and `.conf` are forbidden. Only files at the top level of the `tasks` directory matching the task name regular expression are used; all other files are ignored.
```
DavidS: Are there any other constraints on the `.<ext>`, other than "not `.md` and `.conf`"?

adreyer: not .json

tasks/README.md
```diff
@@ -80,7 +79,11 @@ The preferred style of the keys should be `snake_case`.
       "description": "Description of optional param3",
       "type": "Optional[Integer]"
     }
-  }
+  },
+  "implementations" : [
+    {"name": "foo.sh", "requirements": ["shell"]},
+    {"name": "foo.ps1", "requirements": ["powershell"]}
```
ripienaar: is this a case of first to match for a machine will be used? I imagine there could be overlap in requirements.  Do you have some kind of most suitable situation or just first to match?

Does `puppet-agent` imply that you might require its ruby? should we be concerned with versions?

In MCollective we found it necessary to set requirements with versions because in the long run even the basic thing like `puppet-agent` might change, say a task requires ruby features only available in the ruby provided by agent 6 and not 5 that seems like a real situation we might hit.

Is it really a case of tasks being able to just specify any old stuff here and a runner could decide to make sense of it or not? That seems prone to simple error - I accidentally set `puppet-agetn` as a requirement and it wont complain it will just silently not do what I intend exactly? This free form stuff makes me worry about early user feedback on error and discoverability of errors

tasks/README.md
```diff
@@ -98,15 +101,23 @@ Any parameter that does not specify a type will accept any value (default type i
 
 If a parameter type accepts `null` the task runner will accept either a `null` value or the absence of the property. Task authors must accept missing properties for nullable parameters. Task authors must not differentiate between absent properties and properties with `null` values.
 
+### Metaparameters
+
+In addition to the tasks parameters, the task runner may inject metaparameters prefixed by '_'. These include `_noop` and `_task`.
```
ripienaar: Is `_task` just the name of the current task?

tasks/task.json
```diff
@@ -26,6 +26,24 @@
...
+            "type": "string"
+          },
+          "requirements": {
+            "type": "array",
+            "additionalItems": {
+              "type": "string"
+            }
+          }
+        }
+      }
```
ripienaar: great but i really do wish we had versioned schemas.

adreyer: We're discussing adding a revision number to the task spec that would cover that.

tasks/README.md
adreyer: correct. You can have the same file implement multiple tasks and behave slightly differently.

for example
package::install
package::status
package::uninstall
could all use the package script

tasks/README.md
ripienaar: @adreyer thanks, any comment about the free form nature of these?

adreyer: It seems like that would be quickly apparent. We can probably make it more so by including what features were present when no suitable implementation is found. Generating a warning in pdk seems feasible too

PR #115, opened 2018-06-28
(BOLT-556) Shareable tasks specification

donoghuc:
Update specification for multi-file task support.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -11,7 +11,7 @@ The Puppet Task ecosystem allows users to execute actions on target systems. A *
 
 **Task**: A task is a file and an optional metadata file that will be executed on a target.
 
-**task runner**: The application which executes tasks on target systems. This is the interface for controlling and running tasks, and is assumed to provide some sort of programatic interface (API). In initial releases this is Orchestrator (executing via PXP) and [Bolt] (via ssh/winrm). In the future new task runners may be added. For example the puppet agent may run tasks from resources.
+**task runner**: The application which executes tasks on target systems. This is the interface for controlling and running tasks, and is assumed to provide some sort of programmatic interface (API). In initial releases this is Orchestrator (executing via PXP) and [Bolt] (via ssh/winrm). In the future new task runners may be added. For example the puppet agent may run tasks from resources.
 
 **UI**: User interface, such as CLI tools ([Bolt], `puppet task`) or the Puppet Enterprise Console.
 
@@ -59,9 +59,13 @@ The preferred style of the keys should be `snake_case`.
 
 **parameters**:  The parameters or input the task accepts listed with a [Puppet data type](../language/types_values_variables.md) string and optional description. Top level params names have the same restrictions as Puppet class param names and must match `\A[a-z][a-z0-9_]*\z`. Parameters may be `sensitive` in which case the task runner should hide their values in UI where possible.
 
-**implementations**: A list of implementation files in preference order, along with an optional list of required features for that implementation to be suitable on a target. The available features are defined by the task runner, but task runners should define at least the `shell`, `powershell` and `puppet-agent` features.
+**files**: A list of file resources to be made available to the task executable on the target specified as file paths. Files must be saved in module directories that Puppet makes available via mount points: `files`, `lib`, `tasks`. File specifications ending with `/` will require the entire directory. 
 
-#### Example
+**implementations**: A list of implementation objects. An implementation object describes resources and feature requirements that must be available on a target for specified resources to be utilized. The available features are defined by the task runner; task runners should define at least the `shell`, `powershell` and `puppet-agent` features.
+
+An implementation object has three possible keys `name`, `requirements`, and `files`. The required `name` key is the filename of the task executable. The optional `requirements` key specifies the features that must be present on the target system in order for the task to be run. The optional `files` key specifies file resources to be made available to the task. When `files` are described using both the `files` and `implementations` metadata options the values are joined.   
+
+#### Example Task Metadata
 
 ```json
 {
@@ -83,8 +87,9 @@ The preferred style of the keys should be `snake_case`.
   },
   "implementations" : [
     {"name": "foo.sh", "requirements": ["shell"]},
-    {"name": "foo.ps1", "requirements": ["powershell"]}
-  ]
+    {"name": "foo.ps1", "requirements": ["powershell"], "files": ["my_util/files/util/win_tools.ps1"]}
+  ],
+  "files": ["my_module/lib/puppet_x/helpful_code.rb", "kitchensink/files/task_helpers/"]
 }
 ```
 
@@ -104,8 +109,88 @@ If a parameter type accepts `null` the task runner will accept either a `null` v
 
 ### Metaparameters
 
-In addition to the tasks parameters, the task runner may inject metaparameters prefixed by '_'. These include `_noop` and `_task`.
+In addition to the tasks parameters, the task runner may inject metaparameters prefixed by `_`.
 
+**_noop**: Used to implement logic in tasks to support cases where task should not perform certain actions.
+
+**_task**: Allow multiple task implementations to access the same executable file. The `_task` metaparameter provides the executable the task name to allow task specific logic to be implemented. 
+
+**_installdir**: Tasks with `files` specified in the metadata will be passed the `_installdir` metaparameter to provide the file path to the expected resources. 
+
+#### Metaparameter Examples
+
+When the task runner runs a task with `files` metadata it copies the specified files into a temporary directory on the target. The directory structure of the specified file resources will be preserved such that paths specified with the `files` metadata option will be available to tasks prefixed with `_installdir`. 
+
+##### Python Example
+###### Metadata
+```json
+{
+  "files": ["multi_task/files/py_util/py_helper.py"]
+}
+```
+###### File Resource
+`multi_task/files/py_util/py_helper.py`
+```python
+def useful_python():
+  return dict(helper="python")
+```
+###### Task
+```python
+#!/usr/bin/env python
+import sys
+import os
+import json
+
+params = json.load(sys.stdin)
+sys.path.append(os.path.join(params['_installdir'], 'python_helpers/files'))
+import py_helper
+
+print(json.dumps(py_helper.useful_python()))
+```
+###### Output
+```
+Started on localhost...
+Finished on localhost:
+  {
+    "helper": "python"
+  }
+Successful on 1 node: localhost
+Ran on 1 node in 0.12 seconds
+```
+##### Ruby Example
+###### Metadata
+```json
+{
+  "files": ["multi_task/files/rb_util/rb_helper.rb"]
+}
+```
+###### File Resource
+`multi_task/files/rb_util/rb_helper.rb`
+```ruby
+def useful_ruby
+  { helper: "ruby" }
+end
+```
+###### Task
+```ruby
+#!/usr/bin/env ruby
+require 'json'
+
+params = JSON.parse(STDIN.read)
+require_relative File.join(params['_installdir'], '/ruby_helpers/files/rb_helper.rb')
+
+puts useful_ruby.to_json
+```
+###### Output
+```
+Started on localhost...
+Finished on localhost:
+  {
+    "helper": "ruby"
+  }
+Successful on 1 node: localhost
+Ran on 1 node in 0.12 seconds
+```
 ## Task execution
 
 If the task has multiple implementation files, the `implementations` field of the metadata is used to determine which implementation is suitable for the target. Each implementation can specify `requirements`, which is an array of the required "features" to use that implementation.
@@ -113,6 +198,8 @@ If the task has multiple implementation files, the `implementations` field of th
 If the task has a single implementation file and doesn't use the `implementations` field, that implementation will be used on every target.
 
 The task implementation is copied to the target and then executed on the target by the task runner.
+
+If `files` metadata has been provided those executable resources will be copied to the target and made available to the task via the `_installdir` metaparameter. 
 
 No arguments are passed to the task when it is executed.
 
--- a/tasks/task.json
+++ b/tasks/task.json
@@ -34,16 +34,33 @@
         "required": ["name"],
         "properties": {
           "name": {
-            "type": "string"
+            "type": "string",
+            "description": "Name of task executable file"
           },
           "requirements": {
             "type": "array",
             "additionalItems": {
               "type": "string"
-            }
+            },
+            "description": "Features required on target to execute task"
+          },
+          "files": {
+            "type": "array",
+            "additionalItems": {
+              "type": "string"
+            },
+            "description": "File resources required by task"
           }
         }
-      }
+      },
+      "description": "Rules for selecting implementation resources based on features available on target"
+    },
+    "files": {
+      "type": "array",
+      "items": {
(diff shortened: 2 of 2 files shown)


tasks/README.md
```diff
@@ -61,6 +61,8 @@ The preferred style of the keys should be `snake_case`.
 
 **implementations**: A list of implementation files in preference order, along with an optional list of required features for that implementation to be suitable on a target. The available features are defined by the task runner, but task runners should define at least the `shell`, `powershell` and `puppet-agent` features.
 
+**files**: Array of file names specified as strings that include the module containing the file. File specifications inding with / will require the entire deirector. Files can also be provided as `implementations` objects.
```
ripienaar: `directory`

donoghuc: thanks

tasks/README.md
```diff
@@ -61,6 +61,8 @@ The preferred style of the keys should be `snake_case`.
 
 **implementations**: A list of implementation files in preference order, along with an optional list of required features for that implementation to be suitable on a target. The available features are defined by the task runner, but task runners should define at least the `shell`, `powershell` and `puppet-agent` features.
 
+**files**: Array of file names specified as strings that include the module containing the file. File specifications inding with / will require the entire directory. Files can also be provided as `implementations` objects.
```
MikaelSmith: `inding` -> `ending. I'd also quote the slash `/`.

tasks/README.md
```diff
@@ -84,7 +86,8 @@ The preferred style of the keys should be `snake_case`.
   "implementations" : [
     {"name": "foo.sh", "requirements": ["shell"]},
     {"name": "foo.ps1", "requirements": ["powershell"]}
-  ]
+  ],
+  "files": ["my_module/lib/puppet_x/helpful_code.rb", "my_util/file/util/"]
```
MikaelSmith: We should use `files` instead of `file`, as `files` is the directory actually available as a mount point.

tasks/README.md
MikaelSmith: We should clarify that only module directories that Puppet makes available via mount points will be accessible: files, lib, tasks.

tasks/README.md
```diff
@@ -104,8 +107,13 @@ If a parameter type accepts `null` the task runner will accept either a `null` v
 
 ### Metaparameters
 
-In addition to the tasks parameters, the task runner may inject metaparameters prefixed by '_'. These include `_noop` and `_task`.
+In addition to the tasks parameters, the task runner may inject metaparameters prefixed by `_`.
+
+**_noop**: Used to implement logic in tasks to support cases where task should not perform certain actions.
 
+**_task**: Allow multiple task implementations to access the same executable file. The `_task` metaparameter provides the executable the task name to allow task specific logic to be implemented.
```
MikaelSmith: Personally I'd like this to exclude the module name, i.e. `package::status` would define `_task=status`, `package` would define `_task=init`. We weren't particularly clear before, and having to include the module name in the comparison (when it will always be the same module name) is annoying. This would imply a change to Bolt's existing behavior.

tasks/task.json
```diff
@@ -44,6 +44,18 @@
           }
         }
       }
+    },
+    "files": {
```
MikaelSmith: Need to also update the `implementations` object properties.

tasks/README.md
```diff
@@ -104,8 +107,13 @@ If a parameter type accepts `null` the task runner will accept either a `null` v
...
 ### Metaparameters
 
-In addition to the tasks parameters, the task runner may inject metaparameters prefixed by '_'. These include `_noop` and `_task`.
+In addition to the tasks parameters, the task runner may inject metaparameters prefixed by `_`.
+
+**_noop**: Used to implement logic in tasks to support cases where task should not perform certain actions.
 
+**_task**: Allow multiple task implementations to access the same executable file. The `_task` metaparameter provides the executable the task name to allow task specific logic to be implemented. 
+
+**_installdir**: Tasks with `files` specified in the metatdata will be passed the `_installdir` metaparam to provide the file path to the expected executables.
```
MikaelSmith: I'd like to see an example of using that, to clarify what full path including `_installdir` would look like for some example paths. Do files get flattened? Do they preserve some directory structure? What does it look like?

tasks/README.md
donoghuc: The files as implementation objects will require some more clarification. I imagine a use case for adding them in the implementations section would be to use the "requirement" key for making decisions based on what resources are available on the target. Will work on an example to illustrate.

tasks/task.json
donoghuc: Not exactly sure how to express the expected "items" `implementations` expects. Tried to express as an array of the two possible hash objects.
 {"name": "foo.ps1", "requirements": ["powershell"]}
 {"files": "my_util/files/util/win_tools.ps1", "requirements": ["powershell"]}

tasks/README.md
donoghuc: + 1 for this. Anybody think of a reason not to do it this way?

tasks/README.md
```diff
@@ -84,7 +108,9 @@ The preferred style of the keys should be `snake_case`.
   "implementations" : [
     {"name": "foo.sh", "requirements": ["shell"]},
     {"name": "foo.ps1", "requirements": ["powershell"]}
-  ]
+    {"files": "my_util/files/util/win_tools.ps1", "requirements": ["powershell"]}
```
MikaelSmith: `name` will always be required, it's the file that we run as a task. `files` is an optional property that specifies additional files to copy over.

Note that `requirements` is also optional, lack of it means we'll use it anywhere.

tasks/README.md
```diff
@@ -59,9 +59,33 @@ The preferred style of the keys should be `snake_case`.
 
 **parameters**:  The parameters or input the task accepts listed with a [Puppet data type](../language/types_values_variables.md) string and optional description. Top level params names have the same restrictions as Puppet class param names and must match `\A[a-z][a-z0-9_]*\z`. Parameters may be `sensitive` in which case the task runner should hide their values in UI where possible.
 
-**implementations**: A list of implementation files in preference order, along with an optional list of required features for that implementation to be suitable on a target. The available features are defined by the task runner, but task runners should define at least the `shell`, `powershell` and `puppet-agent` features.
+**files**: Array of file names specified as strings that include the module containing the file. Files must be saved in module directories that Puppet makes available via mount points: files, lib, tasks. File specifications ending with `/` will require the entire directory. Files can also be provided as `implementations` objects.
```
MikaelSmith: Please backtick around `files`, `lib`, and `tasks` to highlight that they're literal paths.

tasks/README.md
```diff
@@ -59,9 +59,33 @@ The preferred style of the keys should be `snake_case`.
...
+**Value**: executable name  
+**Type**: string  
+**Required**: True  
+
+**Key**: "requirements"  
+**Value**: required feature  
+**Type**: string  
+**Required**: False  
+
+`files` can be specified in `implementations` using the `files` key. When `files` are described using both the `files` and `implementations` metadata options the values are joined.
```
MikaelSmith: These sentences are right, but the specification below is wrong. There are 3 properties of an implementations object: `name` (required), `requirements` (optional), `files` (optional).

Also, this format of defining them is very verbose, I'd like something more compact.

tasks/README.md
```diff
@@ -104,8 +130,45 @@ If a parameter type accepts `null` the task runner will accept either a `null` v
...
+exec("#{params['_installdir']}/mymodule/files/script1.py", "--flag", params["flag1"], "-f", params["flag2"], params[arg1])
+```
+##### Using relative path
+###### Metadata
+```json
+{ "files": "task_helpers/files/ruby.rb" }
+```
+###### Task
+```ruby
+require_relative "../../task_helpers/files/ruby.rb"
```
MikaelSmith: I'm not sure I'd include a relative example, I think we want more freedom around how to structure the directory where we copy over files.

tasks/task.json
```diff
@@ -43,6 +43,33 @@
             }
           }
         }
+      },{ 
+        "type": "object",
```
MikaelSmith: Again this is wrong, `files` should be added as an optional property at line 35.

tasks/README.md
MikaelSmith: Supporting why I prefer a single object type: implementations are meant to be selected based on their requirements, the first object where the target matches all its requirements is selected. If we expect to match multiple objects with the same set of requirements, then walking this list gets a lot harder, and the use cases for what that's helpful are unclear to me. Generally if you have extra library files, they go with specific tasks, so it makes sense to group them together.

tasks/README.md
```diff
@@ -59,9 +59,13 @@ The preferred style of the keys should be `snake_case`.
 
 **parameters**:  The parameters or input the task accepts listed with a [Puppet data type](../language/types_values_variables.md) string and optional description. Top level params names have the same restrictions as Puppet class param names and must match `\A[a-z][a-z0-9_]*\z`. Parameters may be `sensitive` in which case the task runner should hide their values in UI where possible.
 
-**implementations**: A list of implementation files in preference order, along with an optional list of required features for that implementation to be suitable on a target. The available features are defined by the task runner, but task runners should define at least the `shell`, `powershell` and `puppet-agent` features.
+**files**: Array of file names specified as strings that include the module containing the file. Files must be saved in module directories that Puppet makes available via mount points: `files`, `lib`, `tasks`. File specifications ending with `/` will require the entire directory. Files can also be provided as `implementations` objects.
```
DavidS: The examples below help, but it took me some back-and-forth here to understand that the `files` here are built like the `source` for puppet `file` resources. Confirming this, or explaining the "mount points" (either directly here, or by reference to pupet documentation) avoid the confusion.

tasks/README.md
```diff
@@ -104,8 +109,28 @@ If a parameter type accepts `null` the task runner will accept either a `null` v
...
+
+**_noop**: Used to implement logic in tasks to support cases where task should not perform certain actions.
+
+**_task**: Allow multiple task implementations to access the same executable file. The `_task` metaparameter provides the executable the task name to allow task specific logic to be implemented. 
+
+**_installdir**: Tasks with `files` specified in the metatdata will be passed the `_installdir` metaparam to provide the file path to the expected executables. 
 
+#### Metaparameter Example
+
+When the task runner runs a task with files it copies those files into the same temporary directory as the task. The task can then load the code based on the path. The directory structure of the specified file resources will be preserved such that paths specified with the `files` metatdata option will be available to tasks prefixed with `_installdir`.
```
donoghuc: For this example we assume that the directory structure is preserved when resource is provided to task via the `_installdir` metaparameter. Do we want to consider other options such as flattening directory structure when copying `files` resources to targets?

tasks/README.md
MikaelSmith: They're not really like `puppet:///` paths, they're full file paths starting from the module name.

tasks/README.md
```diff
@@ -59,9 +59,13 @@ The preferred style of the keys should be `snake_case`.
...
 
-**implementations**: A list of implementation files in preference order, along with an optional list of required features for that implementation to be suitable on a target. The available features are defined by the task runner, but task runners should define at least the `shell`, `powershell` and `puppet-agent` features.
+**files**: Array of file names specified as strings that include the module containing the file. Files must be saved in module directories that Puppet makes available via mount points: `files`, `lib`, `tasks`. File specifications ending with `/` will require the entire directory. Files can also be provided as `implementations` objects. 
 
-#### Example
+**implementations**: A list of implementation objects. An implementation object describes resources and feature requirements that must be available on a target for specified resources to be utilized. The available features are defined by the task runner,  task runners should define at least the `shell`, `powershell` and `puppet-agent` features. An implementation o
```
MikaelSmith: Last 3 words are a fragment. Also remove the double space, and use a semicolon to separate 'features are defined by the task runner', and 'task runners should define'.

tasks/README.md
MikaelSmith: Mentioning that 'Files can also be provided as `implementations` objects.' is not quite correct, and I believe redundant with the `files` section. I'd remove it.

tasks/README.md
MikaelSmith: No, we should always preserve the directory structure.

tasks/task.json
```diff
@@ -41,6 +41,12 @@
             "additionalItems": {
               "type": "string"
             }
+          },
+          "files": {
```
MikaelSmith: We also need this outside the `implementations` array of objects.

Please test out the schema and examples together using something like https://www.jsonschemavalidator.net/.

tasks/README.md
```diff
@@ -59,9 +59,13 @@ The preferred style of the keys should be `snake_case`.
 
 **parameters**:  The parameters or input the task accepts listed with a [Puppet data type](../language/types_values_variables.md) string and optional description. Top level params names have the same restrictions as Puppet class param names and must match `\A[a-z][a-z0-9_]*\z`. Parameters may be `sensitive` in which case the task runner should hide their values in UI where possible.
 
-**implementations**: A list of implementation files in preference order, along with an optional list of required features for that implementation to be suitable on a target. The available features are defined by the task runner, but task runners should define at least the `shell`, `powershell` and `puppet-agent` features.
+**files**: Array of file names specified as strings that include the module containing the file. Files must be saved in module directories that Puppet makes available via mount points: `files`, `lib`, `tasks`. File specifications ending with `/` will require the entire directory.
```
donoghuc: Do we prefer "array" or "list" for the description? I noticed for `implementations` "list" was used. In the JSON schema the type is "array".

MikaelSmith: I don't believe JSON schema has a type that is "list". But we tend to use 'list' and 'array' interchangeably in description, and list is a bit more colloquial.

tasks/README.md
```diff
@@ -104,8 +109,28 @@ If a parameter type accepts `null` the task runner will accept either a `null` v
...
+
+**_task**: Allow multiple task implementations to access the same executable file. The `_task` metaparameter provides the executable the task name to allow task specific logic to be implemented. 
+
+**_installdir**: Tasks with `files` specified in the metadata will be passed the `_installdir` metaparameter to provide the file path to the expected resources. 
 
+#### Metaparameter Example
+
+When the task runner runs a task with `files` metadata it copies the specified files into a temporary directory on the target. The directory structure of the specified file resources will be preserved such that paths specified with the `files` metadata option will be available to tasks prefixed with `_installdir`. 
+
+###### Metadata
```
MikaelSmith: Let's add a few more examples around using libraries here. Ruby might look like
```
require 'json'
params = JSON.parse(STDIN.read)
$LOAD_PATH.unshift(File.join(params['_installdir'], 'lib'))
require 'puppetx'
...
```

tasks/README.md
```diff
@@ -104,8 +109,28 @@ If a parameter type accepts `null` the task runner will accept either a `null` v
...
+```json
+{ "files": ["mymodule/files/script1.py"] }
+```
+###### Task
+```python
+require json
```
MikaelSmith: This isn't actual Python, it's Ruby written in a Python style. Should be something like
```
import json, sys, os
params = json.load(sys.stdin)
os.execl([params['_installdir']+'/mymodule/files/script1.py', '--flag', params['flag1'], '-f', params['flag2'], params['arg1']])
```

MikaelSmith:
Thanks for your feedback! We're starting to plan out how to implement this in Bolt and PE now.

PR #119, opened 2018-08-07
(maint) Add revision, private and extensions to task-spec

adreyer:


adreyer:
I question whether the file work should be rev 3 or rev 4.

tasks/README.md
```diff
@@ -107,19 +112,39 @@ Any parameter that does not specify a type will accept any value (default type i
 
 If a parameter type accepts `null` the task runner will accept either a `null` value or the absence of the property. Task authors must accept missing properties for nullable parameters. Task authors must not differentiate between absent properties and properties with `null` values.
 
+### Options
+
+**type**: A Puppet Type string describing the type of value the parameter accepts. Default is `Any`. (rev 1)
+
+**description**: A string description of the parameter. (rev 1)
```
MikaelSmith: This should call out `sensitive` as well.

tasks/README.md
```diff
@@ -107,19 +112,39 @@ Any parameter that does not specify a type will accept any value (default type i
...
+
+## Task implementations
+
+The implementations property is used to describe different implementations for differet targets. The value is an array of implementation options objects. For each target the task is run on the Task Runner must compare the required features of the target to it's list of available features on that target and execute the first implementation where all requirements are statisfied.
+
+### Options
+
+**name**: The file name of the task file that contains the implementation of the task. This must be at the top level of the tasks directory of the module.(rev 2)
```
MikaelSmith: In conjunction with the example you changed above, I think it'd make sense to call out the pattern we recommend to remain compatible with runners implementing revision 1.

tasks/README.md
```diff
@@ -107,19 +112,39 @@ Any parameter that does not specify a type will accept any value (default type i
...
+
+**requirements**: A list of features the target is required to support for the implementation to be used. `shell`, `powershell` and `puppet-agent` features are added by the ssh, winrm, and pcp transports respectively. (rev 2)
+
+**input_method**: The input method to use for this implementation of the task. Default empty [] which will make this implementation suitable for all targets. (rev 3)
```
MikaelSmith: Use back-ticks around `[]`.

tasks/README.md
```diff
@@ -107,19 +112,39 @@ Any parameter that does not specify a type will accept any value (default type i
...
+
 ### Metaparameters
 
 In addition to the tasks parameters, the task runner may inject metaparameters prefixed by `_`.
 
-**_noop**: Used to implement logic in tasks to support cases where task should not perform certain actions.
+**_noop**: Used to implement logic in tasks to support cases where task should not perform certain actions. (rev 1)
 
-**_task**: Allow multiple task implementations to access the same executable file. The `_task` metaparameter provides the executable the task name to allow task specific logic to be implemented. 
+**_task**: Allow multiple task implementations to access the same executable file. The `_task` metaparameter provides the executable the task name to allow task specific logic to be implemented. (rev 3)
```
MikaelSmith: The only distinction I've noticed between rev 2 and rev 3 is that we claim Bolt supports rev 2. So should we put `_task` and `_installdir` as rev 2?

tasks/README.md
```diff
@@ -268,8 +293,8 @@ Task output is consumed on stdout and made available through the task runner API
 If the output cannot be parsed as a JSON object (i.e. `{...}`), one is created by the task runner with the captured stdout stored in the  `_output` key: i.e. `{"_output": "the string of output"}`. Otherwise the parsed JSON object is used.
 
 All '_' prefixed keys are reserved and should only be used as described below:
-- **_output**: A text summary of the job result. If the job does not return a JSON object the contents of stdout will be put in this key.
-- **_error**: Tasks can set this when they fail and the UI can more gracefully display messages from it.
+- **_output**: A text summary of the job result. If the job does not return a JSON object the contents of stdout will be put in this key. (rev 1)
```
MikaelSmith: In these lines you put a space before `(rev N)`, unlike everywhere else. I'd like that to be consistent throughout the doc.

tasks/task.json
```diff
@@ -61,7 +61,17 @@
...
+    "private": {
+      "type": "boolean",
+      "description": "Should this task appear by default in UI lists of tasks"
+    },
+    "extensions": {
+      "type": "object",
+      "description": "Task Runner specific metadata extensions",
+      "items": {
+        "type": "object"
+      }
```
MikaelSmith: Needs another closing brace `}`.

tasks/README.md
adreyer: _task at least. Bolt doesn't yet support installdir right?

MikaelSmith: Ah, right `_installdir` is not in yet.

donoghuc:
@MikaelSmith I went ahead and updated based on your comments.

PR #122, opened 2018-08-29
(PUP-8635) Add specification of the Deferred data type

hlindberg:


language/types_values_variables.md
```diff
@@ -2185,6 +2188,52 @@ There is no automatic unwrapping of sensitive values. As a consequence it is not
...
+
+Creates a new instance of the `Deferred` type. It accepts the name of the function to later call, and an array of
+arguments to use when making the call. The arguments are evaluated at the time the deferred value is created, but
+any deferred values among the arguments will be resolved as part of the resolution process (when actually making the call).
+
+    $x = Deferred('some_function', ['argument1', 42])
+
+Will, when the deferred is resolved result in the call:
+
+    some_function('argument1', 42)
```
ripienaar: and presumably these would be ruby functions? or pp functions too? Or something else entirely (would they have scope and facts?)

hlindberg: In first release only ruby functions. They are called in a global scope. `$facts` is set with the facts when deferred is resolved on the agent side.
(If resolved while compiling - if you call the deferred, it has everything the compiler has).

hlindberg:
The specification should include a "since puppet 6.0.0" for Deferred.

language/types_values_variables.md
ripienaar: nice, ok thanks

hlindberg: PR updated with more details

PR #128, opened 2018-09-26
(SPEC) Add positional arguments to the task spec

davejohnston:
The objective of this change is to make integrating with existing
scripts easier.  i.e. rather than having to write an adapter to read
json from stdin, or read environment variables, the user can just
specify the metadata necessary to describe and execute their script.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -246,7 +246,7 @@ Future releases of the task runner may expand this or add configuration options 
 
 ## Input Methods
 
-There are a three input methods available to tasks: [stdin](#stdin), [environment](#environment-variables), and [powershell](#powershell). By default tasks without a `.ps1` extension are passed parameters with both the `stdin` and `environment` input methods. Tasks with a `.ps1` extension are passed parameters with the `powershell` input method by default.
+There are four input methods available to tasks: [stdin](#stdin), [environment](#environment-variables), [positional](#positional) and [powershell](#powershell). By default tasks without a `.ps1` extension are passed parameters with both the `stdin` and `environment` input methods. Tasks with a `.ps1` extension are passed parameters with the `powershell` input method by default.  Positional parameters are passed to the task in the order specified by the parameter definition.
 
 In the future we may support other formats and methods for passing params to the task.
 
@@ -261,6 +261,9 @@ Params are set as environment variables before the task is executed.
 Param names will be prefixed with `PT_` to create the environment variables in the form `PT_<param_name>` (i.e. the value of the `foobar` param is set as `PT_foobar`).
 
 String parameters will not be quoted. Numerical parameters and structured objects will use their JSON representations.
+
+### Positional 
+Positional input is useful for converting existing scripts to tasks.  The parameters are passed to the task as arguments.  The positional/order they are passed should be defined by the parameter.   If the position has not been specified then each parameter will be passed to the task in a unspecified order.
 
 #### Example
 
@@ -280,6 +283,23 @@ PT_a=1
 PT_b=a string
 PT_c=[1, 2, "3"]
 PT_d={"x": {"y": [0]}}
+```
+
+##### Positional
+```bash
+arg1 = $1
+arg2 = $2
+```
+
+```ps
+[CmdletBinding()]
+ param (
+   [Parameter(Position=0)]
+   [string]$arg1,
+
+   [Parameter(Position=1)]
+   [string]$arg2
+ )
 ```
 
 ### Powershell
--- a/tasks/task.json
+++ b/tasks/task.json
@@ -20,7 +20,7 @@
     },
     "input_method": {
       "type": "string",
-      "enum": ["stdin", "environment"],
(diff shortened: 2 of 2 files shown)


davejohnston:
@MikaelSmith @adreyer I am interested in trying to make it easier for customers to create tasks using their existing scripts (without having to make any modifications or create adapters).   I'm looking at this from a discovery perspective - so proposing the above, whereby we could write a task where the task runner will pass parameters directly to the script in a specified order.   e.g

my_script.sh <param1> <param2> <param3>
or
my_scripts.ps1 <param1> <param2> <param3>

Do you think this is a reasonable thing to include in the spec ?

tasks/README.md
```diff
@@ -246,7 +246,7 @@ Future releases of the task runner may expand this or add configuration options
 
 ## Input Methods
 
-There are a three input methods available to tasks: [stdin](#stdin), [environment](#environment-variables), and [powershell](#powershell). By default tasks without a `.ps1` extension are passed parameters with both the `stdin` and `environment` input methods. Tasks with a `.ps1` extension are passed parameters with the `powershell` input method by default.
+There are four input methods available to tasks: [stdin](#stdin), [environment](#environment-variables), [positional](#positional) and [powershell](#powershell). By default tasks without a `.ps1` extension are passed parameters with both the `stdin` and `environment` input methods. Tasks with a `.ps1` extension are passed parameters with the `powershell` input method by default.  Positional parameters are passed to the task in the order specified by the parameter definition.
```
hlindberg: The parameter metadata is in a JSON Object, and it is therefore not guaranteed that the read result is always in the order keys were written. With a Ruby runtime >= 1.9 the result is a hash with the keys in expected order, but this is a Ruby JSON implementation detail. To correctly specify the order I suggest either adding a `position` argument to a parameter definition, or to add a separate array with the parameter names in the wanted order.

davejohnston: @hlindberg thanks - I have a position argument defined here: https://github.com/puppetlabs/puppet-specifications/pull/128/files#diff-f2efe900dbbc251d3e29d4e5fa32c701R104  Does that resolve your concern?   Or were you intending something else ?

tasks/task.json
```diff
@@ -100,6 +100,10 @@
             "description": "Whether the task runner should treat the parameter value as sensitive",
             "type": "boolean",
             "default": false
+          },
+          "position": {
+            "description": "The position the parameter should be passed to the script",
```
MikaelSmith: Does this start with 0 or 1 indexing? Is it an error if sequential parameters are not declared in metadata, or provided when running the command? Are all positional parameters required?

tasks/README.md
```diff
@@ -262,6 +262,9 @@ Param names will be prefixed with `PT_` to create the environment variables in t
 
 String parameters will not be quoted. Numerical parameters and structured objects will use their JSON representations.
 
+### Positional 
+Positional input is useful for converting existing scripts to tasks.  The parameters are passed to the task as arguments.  The positional/order they are passed should be defined by the parameter.   If the position has not been specified then each parameter will be passed to the task in a unspecified order.
```
MikaelSmith: The `positional` property should be added under the `Task Paremeters` section. We should also mark this as part of `rev 4` of the spec.

tasks/task.json
MikaelSmith: I'd put those details in the Task Parameters > Options section when describing the `position` property.

davejohnston:
@MikaelSmith thanks - some good questions/comments.  I'm going to get our current release out of the way and then come back and maybe create some worked examples for this,

tasks/task.json
hlindberg: To be flexible the scheme could output parameters sorted on position/order. That means it could use 10, 20, 30 etc to allow for parameters to be inserted in the future - i.e. less
busy work if another positional parameter needs to be injected somewhere in the middle.
The specification also needs to define what happens if some parameters have position/order and others don't. For example; positional parameters are given first (in order) and the remaining parameters are output in undefined order.

adreyer:
This was discussed as part of the original task spec. We ended up cutting it for a few of reasons:

1. Most scripts need flags not positional arguments. Multiple positional arguments lead to bad scripts
2. Proper escaping is difficult and results in user confusion when things break.
3. Bolt can run scripts with positional arguments. If you have a script and don't want to turn it into a task just run the script. This format has better support for flags too.
4. In my experience converting a script from flags or positional arguments takes <30m. It's pretty easy to do and results in a higher quality task long term.
Since then:
5. The new multi-file support for bolt means that you can write a simple wrapper for you script, put the script in `files` then execute the script with the wrapper. This decreases the time to convert a script to as task to <10m and solves the use case where the script is provided by a 3rd party and you don't want to. I've ticketed better documenting this process in BOLT-989.

I'm going to close this ticket.

@davejohnston if the underlying issue is the ability of discovery users to run preexisting scripts I think we should discuss just running scripts directly with bolt-server or boltshim rather then forcing users to convert those into tasks.

PR #131, opened 2018-11-20
(BOLT-791) Add remote tasks to spec

adreyer:


tasks/README.md
```diff
@@ -144,6 +148,8 @@ In addition to the tasks parameters, the task runner may inject metaparameters p
 
 **_installdir**: Tasks with `files` specified in the metadata will be passed the `_installdir` metaparameter to provide the file path to the expected resources.(rev 3)
 
+**_target**: Connection information for connectin to the real target when the task is running on a proxy.
```
MikaelSmith: `connectin` -> `connecting`

tasks/README.md
```diff
@@ -347,3 +353,14 @@ Some examples:
 The JSON schema for errors is included in [error.json](error.json).
 
 [Bolt]: https://github.com/puppetlabs/bolt
+
+## Exectuting Remote Tasks
```
MikaelSmith: `Exectuting` -> `Executing`

tasks/README.md
MikaelSmith: Should probably include the criteria that cause it to be included.

tasks/README.md
```diff
@@ -347,3 +353,14 @@ Some examples:
...
+
+Tasks may be written to execute on a proxy target and interact remotely with
+the specified target. This is useful when the target has a limited shell
+environment or only exposes an HTTP API.
+
+- The task runner must refuse to execute tasks implementations that do not have `remote` set on remote targets.
+- The task runner must refuse to execute task implementations that do have `remote` set on normal targets.
```
DavidS: An implementation that changes behaviour depending on the presence of the `_target` metaparam is easily imaginable. Are we making this too strict by disallowing a task to provide a single implementation for both local and remote execution?

tasks/README.md
```diff
@@ -69,6 +69,8 @@ The preferred style of the keys should be `snake_case`.
 
 **extensions**: A hash of extensions to the task spec used by a Specific Task Runner. Each key at the top level should be the name of the Task Runner the extension is used by. Task Runners should not read extensions outside of their own namespace.(rev 3)
 
+**remote**: Default `false`. All implementation of this task operate on a remote target using the `target` metaparam.
```
DavidS: does this require a `rev 4` annotation? (same question also on for the same change below)

tasks/README.md
```diff
@@ -347,3 +353,14 @@ Some examples:
 The JSON schema for errors is included in [error.json](error.json).
 
 [Bolt]: https://github.com/puppetlabs/bolt
+
+## Exectuting Remote Tasks
+
+Tasks may be written to execute on a proxy target and interact remotely with
+the specified target. This is useful when the target has a limited shell
+environment or only exposes an HTTP API.
```
DavidS: strike "HTTP" before "API". It is a unnecessary restriction.

tasks/README.md
DavidS: `target` -> `_target`

tasks/README.md
DavidS: Do we want to carry a schema here similar to task parameters?

tasks/README.md
adreyer: This is accomplished but having two entries in `implementations` that point to same file, one remote the other not.

tasks/README.md
adreyer: yes, I stopped adding `rev 4` so I wouldn't have to change if it something else took 4.

tasks/README.md
adreyer: sending the schema down to the task next to the data doesn't seem to accomplish much. we could add a schema to task metadata but that would be painful to use since you'd have to copy it everywhere.  The correct solution for this IMO is to have a {{target-type}} setting in metadata. That would reference puppet device implementation that would own the schema. In the interest of getting this working I'm leaving that as a separate enhancement to the task spec.

tasks/README.md
DavidS: ah, oh. that makes sense.

tasks/README.md
DavidS: oh no, not sending the schema to the task, only having it in the task metadata (like the parameter schema).

When you're pointing at the target-type, I do agree that replicating the schema across tasks is "less than useful".

adreyer: I do think we need some concept of type in the inital iteration of this. We can't support `apply` or `facts` without knowing what provider to use. I think we can do that without supporting schema checking yet.

tasks/README.md
```diff
@@ -347,3 +353,14 @@ Some examples:
...
+
+- The task runner must refuse to execute tasks implementations that do not have `remote` set on remote targets.
```
lucywyman: s/tasks implementations/task implementations/

PR #133, opened 2018-12-04
WIP - (FM-7596) Draft of RSAPI Transport layer

da-ar:
Not happy with the transport implementation being described here, I think we need an interface class.

Added file paths to the code samples, as in @DavidS notes. I think they help make things clearer.

### DIFF
--- a/language/resource-api/README.md
+++ b/language/resource-api/README.md
@@ -378,8 +378,8 @@ To support Puppet versions prior to 6, see the [Legacy Support](#legacy-support)
 # lib/puppet/transport/nexus_schema.rb
 Puppet::ResourceAPI.register_transport(
   name: 'nexus', # points at class Puppet::Transport::Nexus
-  desc: 'Connects to a Cisco Nexus device'
-  features: [], # TODO
+  desc: 'Connects to a Cisco Nexus device',
+  # features: [], # future extension points
   connection_info: {
     connector_type: {
       type: 'Enum[nxapi, ssh, grpc, pidgeon]',
@@ -450,7 +450,7 @@ To allow implementors a wide latitude in implementing connection and retry handl
 
 1. `verify` and `facts`, like `initialize`, only needs to throw exceptions when unrecoverable errors are encountered, or when the retry logic times out.
 
-1. On any other operation will primarily be used by providers and tasks talking to this kind of target. Those operations are free to use any error signalling mechanism that is convenient. Providers and tasks will have to provide appropriate error signalling to the runtime.
+1. Any other methods on the transport are to be used by providers and tasks talking to this kind of target. Those operations are free to use any error signalling mechanism that is convenient. Providers and tasks will have to provide appropriate error signalling to the runtime.
 
 ### Direct Access to Transports
 
@@ -471,7 +471,19 @@ To get a list of all registered transports, call the `list` method:
 
 `Puppet::ResourceApi::Transport.list`
 
-It will return a hash of all registered transport schemas keyed by their name. Each entry in the list is the transport schema definition as passed to `register_transport`.
+It will return a hash of all registered transport schemas keyed by their name. Each entry in the list is the transport schema definition as passed to `register_transport`:
+
+
+```ruby
+{
+  'nexus' => {
+    name: 'nexus',
+    desc: 'Connects to a Cisco Nexus device',
+    # ...
+    # the rest of the transport schema definition here
+  },
+}
+```
 
 ### Legacy Support
 
@@ -503,13 +515,14 @@ Inheriting from `Puppet::ResourceApi::Transport::Wrapper` will ensure that the n
 
 ### Porting existing code
 
-To port your existing device code as a transport, move the class to `Puppet::Transport`, remove mention of the `Puppet::Util::NetworkDevice::Simple::Device` (if it was used) and change the initialisiation to accept and process a `connection_info` hash instead of the previous structures.
+To port your existing device code as a transport, move the class to `Puppet::Transport`, remove mention of the `Puppet::Util::NetworkDevice::Simple::Device` (if it was used) and change the initialisation to accept and process a `connection_info` hash instead of the previous structures. When accessing the `connection_info` in your new transport, change all string keys to symbols (e.g. `'foo'` to `:foo`). Add `context` as first argument to your `initialize` and `facts` method.
 
 The following replacements have provided a good starting point for these changes:
 
 * `Util::NetworkDevice::NAME::Device` -> `ResourceApi::Transport::NAME`
 * `puppet/util/network_device/NAME/device` -> `puppet/transport/NAME`
 * `device` -> `transport`
+* Replace all direct calls to puppet logging with calls into the `context` logger, potentially getting the context instance from the provider.
 
 Of course this list can't be exhaustive and will depend on the specifics of your codebase.
 
--- a/language/resource-api/README.md
+++ b/language/resource-api/README.md
@@ -368,9 +368,9 @@ class Puppet::Provider::Nx9k_vlan::Nexus
   end
 ```
 
-Declaring this feature restricts the resource from being run "locally". It is expected to execute all external interactions through the `context.transport` instance. The way that instance is set up is runtime specific. In Puppet, it is configured through the [`device.conf`](https://puppet.com/docs/puppet/5.3/config_file_device.html) file, and only available when running under [`puppet device`](https://puppet.com/docs/puppet/5.3/man/device.html).
+Declaring this feature restricts the resource from being run "locally". It executes all external interactions through the `context.transport` instance. The way the instance is set up is runtime specific. In Puppet, it is configured through the [`device.conf`](https://puppet.com/docs/puppet/5.3/config_file_device.html) file, and only available when running under [`puppet device`](https://puppet.com/docs/puppet/5.3/man/device.html).
 
-To support puppet versions prior to 6, please see the [Legacy Support](#legacy-support) section below.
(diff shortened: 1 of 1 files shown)


language/resource-api/README.md
```diff
@@ -333,33 +344,101 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+# lib/puppet/provider/nx9k_vlan/nexus.rb
+class Puppet::Provider::Nx9k_vlan::Nexus
   def set(context, changes, noop: false)
     changes.each do |name, change|
       is = change.has_key? :is ? change[:is] : get_single(name)
       should = change[:should]
       # ...
-      context.device.do_something unless noop
+      if context.transport.verify
+        context.transport.do_something if noop
```
Thomas-Franklin: Should this not be `unless noop`?

da-ar: lol - yep

language/resource-api/README.md
```diff
@@ -333,33 +344,101 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+    hostname: {
+      type: 'String',
+      desc: 'The host to connect to.',
+    },
+    username: {
+      type: 'String',
+      desc: 'The user.',
+    },
+    password: {
+      type: 'Sensitive[String]',
```
adreyer: We discussed using sensitive like this in the task spec and decided it wasn't what we wanted. This means the password field will only accept a value of type Senstitive where what we mean is this value should be treated as sensitive information. For the task spec we instead added a new key `sensitive` next to type and description to capture this. I would encourage doing something similar here.

language/resource-api/README.md
```diff
@@ -333,33 +344,101 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+  def close; end
+end
+```
+
+The transport layer consists of the schema and the implementation. The schema is defined in the same manner as a `Type`, except instead of `attributes`,
+ there is now `connection_info` and it describes the shape of the data which should be passed to the implementation for a connection to be made.
+
+The implementation must follow the structure as defined above and provide the following methods:
+
+  * initialize(connection_info)
```
adreyer: Is initialize expected to open the connection?

language/resource-api/README.md
```diff
@@ -333,33 +344,101 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+    * `transport_info` contains validated hash matching schema
+  * facts
+    * Access the device and return a facts hash
+  * verify
+    * Perform an operation that returns true if the connection is correctly established
```
adreyer: If the connection could be established but is currenlty closed should verify return false? How would the transport recover from something like a timeout? Is that up to the code using the transport?

language/resource-api/README.md
```diff
@@ -333,33 +344,101 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+  * close
+    * Close the connection
+
+### Legacy Support
+
+Prior support for `remote_resource` was provided through `Puppet::Util::NetworkDevice`. This code can be updated by inheriting from `Puppet::ResourceApi::Transport::Forwarder` rather than `Puppet::Util::NetworkDevice::Simple::Device`.
```
adreyer: Is there a reason we don't just make `Puppet::Util::NetworkDevice::Simple::Device` the forwarder?

language/resource-api/README.md
```diff
@@ -333,33 +344,101 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+#  lib/puppet/transport/nexus.rb
+class Puppet::Transport::Nexus
+  def initialize(transport_info); end
+  def facts; end
+  def verify; end
+  def close; end
+end
+```
+
+The transport layer consists of the schema and the implementation. The schema is defined in the same manner as a `Type`, except instead of `attributes`,
```
DavidS: Put this with an additional sentence what a `transport` is at the top of the chapter

language/resource-api/README.md
adreyer: the run_task function turns them into instances of Puppet::Sensitive and then the transports unwrap the sensitive values at the last minute.

https://github.com/puppetlabs/bolt/blob/fb186f9165103b399a6960a7fc42983c4ce714dd/bolt-modules/boltlib/lib/puppet/functions/run_task.rb#L135
https://github.com/puppetlabs/bolt/blob/fb186f9165103b399a6960a7fc42983c4ce714dd/lib/bolt/transport/ssh.rb#L110

language/resource-api/README.md
adreyer: ah cool this makes more sense.

language/resource-api/README.md
```diff
@@ -333,33 +344,101 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+  * close
+    * Close the connection
+
+### Legacy Support
+
+Prior support for `remote_resource` was provided through `Puppet::Util::NetworkDevice`. This code can be updated by inheriting from `Puppet::ResourceApi::Transport::Forwarder` rather than `Puppet::Util::NetworkDevice::Simple::Device`.
+
+```ruby
+# lib/puppet/util/network_device/nexus/device.rb
+module Puppet::Util::NetworkDevice::Nexus
```
DavidS: This will need a  `require 'puppet/resource_api'`

language/resource-api/README.md
DavidS: Added language explaining this.

language/resource-api/README.md
DavidS: Added more extensive explanation to the doc

language/resource-api/README.md
DavidS: Added more extensive explanation to the doc, and fixed up the error handling to be ruby-style, not go-style.

language/resource-api/README.md
DavidS: Fixed the paragraph.

da-ar:
+1

DavidS:
@clairecadman can you have a quick look over the changes here, and apply your magic, please?

clairecadman:
@DavidS Yes I'll take a look this afternoon!

clairecadman:
@DavidS I made a couple of minor edits. But line 453 needs to be fixed, I'm just not sure what it means. Thanks!

language/resource-api/README.md
```diff
@@ -333,33 +344,166 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+2. `connect` to the transport by name, passing the connection info
+3. When the transport has been initialized, `connect` will return the `Transport` object
+
+See above for possible exceptions coming from initializing a transport.
+
+To get a list of all registered transports, call the `list` method:
+
+`Puppet::ResourceApi::Transport.list`
+
+It will return a hash of all registered transport schemas keyed by their name. Each entry in the list is the transport schema definition as passed to `register_transport`.
```
Magisus: Just to be clear, does this mean each value is also a hash (exactly like the one passed to `register`)? Thinking about the API access to this, and want to figure out what kinds of data conversion we would need (or wouldn't need) to do.

language/resource-api/README.md
```diff
@@ -333,33 +344,166 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+
+```ruby
+# lib/puppet/transport/nexus_schema.rb
+Puppet::ResourceAPI.register_transport(
+  name: 'nexus', # points at class Puppet::Transport::Nexus
+  desc: 'Connects to a Cisco Nexus device'
+  features: [], # TODO
+  connection_info: {
+    connector_type: {
+      type: 'Enum[nxapi, ssh, grpc, pidgeon]',
```
ALawley: Will this be supporting the `Optional` data type?

language/resource-api/README.md
```diff
@@ -333,33 +344,166 @@ When a resource is marked with `noop => true`, either locally or through a globa
...
+
+1. On `initialize`, the transport may apply deeper validation to the passed connection information, like mutual exclusivity of optional values (e.g. `password` OR `key`) and throw an `ArgumentError`
+1. On `initialize`, the transport may (but doesn't have to) try to establish a connection to the remote target. If this fails due to unrecoverable errors (e.g. "password rejected"), an appropriate exception will be thrown.
+
+1. `verify` and `facts`, like `initialize`, only needs to throw exceptions when unrecoverable errors are encountered, or when the retry logic times out.
+
+1. On any other operation will primarily be used by providers and tasks talking to this kind of target. Those operations are free to use any error signalling mechanism that is convenient. Providers and tasks will have to provide appropriate error signalling to the runtime.
```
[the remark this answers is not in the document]
DavidS: Me neither. And I wrote it. :embarrassed: Let me try again:

> Any other methods on the transport are to be used by providers and tasks talking to this kind of target. ...

language/resource-api/README.md
DavidS: Yes. `list` will return a ruby hash of hashes. I'll need to improve the wording on this one a little bit to make that clearer.

The current implementation is quite primitive: https://github.com/puppetlabs/puppet-resource_api/pull/150/commits/bd3a21e0abd825f55ef928189d5ebb8a256b7c6c

language/resource-api/README.md
DavidS: The Resource API does not put any restrictions on the types folks can use here.

That said, I would understand if you want to add guidance to developers around what can be mapped into UI (String (including Pattern and Enum), Number, Variant, Optional, Array, sensitive) and do a generic JSON(?) input field for everything else.

da-ar:
@DavidS reminder that you have some local changes to push.

PR #134, opened 2019-02-14
PE-25678 Task spec versioning

melamos:


tasks/README.md
```diff
@@ -1,4 +1,4 @@
-# Puppet Task Spec (Version 1)
+# Puppet Task Spec (version 1, revision 4)
```
melamos: I wasn't totally sure, but I think we're currently at rev 4?

tasks/README.md
```diff
@@ -17,14 +17,25 @@ The Puppet Task ecosystem allows users to execute actions on target systems. A *
...
+These are the revisions to this version of the task spec. 
+
+**Note**: Bolt always uses the latest version and revision of the task spec. Puppet Enterprise versions use a specific version and revision of the task spec, indicated below. 
+
+| Revision | Changes                  | PE version                |
+|----------|--------------------------|---------------------------|
+| 1        |                          | 2017.3.z 2018.1.z         |
+| 2        | Cross-platform tasks.    | 2019.0.0                  |
+| 3        | Multiple files per task. | 2019.0.1                  |
+| 4        | Hide private tasks.      |                           |
```
melamos: Is rev 4 in 2019.0.2?

tasks/README.md
```diff
@@ -17,14 +17,25 @@ The Puppet Task ecosystem allows users to execute actions on target systems. A *
...
 
 Older versions of the task runner may not support tasks that rely on newer features of the task spec. The task spec version is not intended to capture this now. Authors should use the `puppet_task_version` field in the module's metadata to document such incompatibilities for users.
 
 Task Runners should document which version and revision of the task spec they support.
 
+### Revision history
+
+These are the revisions to this version of the task spec. 
+
+**Note**: Bolt always uses the latest version and revision of the task spec. Puppet Enterprise versions use a specific version and revision of the task spec, indicated below.
```
melamos: Is it true that Bolt (latest) always uses the latest task spec?

tasks/README.md
adreyer: No. I don't know when we'll add the "hide private tasks" feature. It's a bit different from the other ones since it doesn't prevent tasks any tasks from working.

tasks/README.md
adreyer: No we usually merge changes in here before releasing bolt.

melamos: I included the info about PE here because the related JIRA ticket said PS wanted a matrix. But on second thought, I wonder if it would suffice to indicate in each version of PE which task spec is supported, as I suggested in the ticket. That would enable this spec to stand alone, and solve the awkward y.z versioning issues in this table.

PR #139, opened 2019-05-09
(FM-7878) Update transport docs

DavidS:
There were some inaccuracies vs the actual implementation and some copy
editing to make the document flow better.

### DIFF
--- a/language/resource-api/README.md
+++ b/language/resource-api/README.md
@@ -363,11 +363,11 @@ class Puppet::Provider::Nx9k_vlan::Nexus
   end
 ```
 
-Declaring this feature restricts the resource from being run "locally". It executes all external interactions through the `context.transport` instance. The way the instance is set up is runtime specific. In Puppet, it is configured through the [`device.conf`](https://puppet.com/docs/puppet/5.3/config_file_device.html) file, and only available when running under [`puppet device`](https://puppet.com/docs/puppet/5.3/man/device.html).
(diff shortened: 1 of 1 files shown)


language/resource-api/README.md
```diff
@@ -465,42 +481,9 @@ To allow implementors a wide latitude in implementing connection and retry handl
...
-    # ...
-    # the rest of the transport schema definition here
-  },
-}
-```
-
-### Legacy Support
-
-Before Puppet 6.X (TBD), remote resources were only supported through the `Puppet::Util::NetworkDevice` namespace. To make a module useful on these older versions, a shim `Device` class needs to be provided to connect the dots:
+Before the Resource API, remote resources were only supported through the `Puppet::Util::NetworkDevice` namespace. To connect your transport to what Puppet expects, you need to provide a shim `Device` class.
```
clairecadman: Maybe change: "To connect your transport to what Puppet expects, you need to provide a shim `Device` class."

To: "To connect your transport *in a way that* Puppet *understands*, you need to provide a shim `Device` class."

If this is what it means? If not please clarify.

language/resource-api/README.md
```diff
@@ -524,9 +507,11 @@ module Puppet::Util::NetworkDevice::Nexus
 end
 ```
 
-Inheriting from `Puppet::ResourceApi::Transport::Wrapper` will ensure that the necessary `Device` methods will be implemented using your transport.
+Inheriting from `Puppet::ResourceApi::Transport::Wrapper` ensures that the necessary `Device` methods are implemented using your transport. Specify the transport name in the `super` call to make the connection.
+
+> Note that because of the way the Resource API is bundled with Puppet agent packages, agent versions 6.0 through 6.3 are incompatible with this way of executing remote content. These versions will not be supported after [PE 2019.0 in August 2019](https://puppet.com/misc/puppet-enterprise-lifecycle).
```
clairecadman: 2019.0 has been out for a while? This sounds like it comes out in August 2019.

language/resource-api/README.md
DavidS: That sounds so much better.

language/resource-api/README.md
DavidS: changed to `These versions will not be supported after [support for PE 2019.0 ends in August 2019](https://puppet.com/misc/puppet-enterprise-lifecycle).` for clarity

DavidS:
@clairecadman fixed the last two issues. Please merge if you're satisfied.

PR #140, opened 2019-05-24
(MODULES-9428) Clarify composite namevar edgecases and add examples

DavidS:


language/resource-api/README.md
```diff
@@ -208,18 +209,18 @@ class Puppet::Provider::AptKey::AptKey < Puppet::ResourceApi::SimpleProvider
 Once all of that is in place, instead of the `set` method, the provider needs to implement the `create`, `update` or `delete` methods:
 
 * `create(context, name, should)`: This is called when a new resource should be created.
-  * `context`: provides utilties from the runtime environment, and is decribed in more detail there.
-  * `name`: the name or hash of the new resource.
+  * `context`: provides provides utilities from the runtime environment, and is described in more detail there.
```
hlindberg:
```suggestion
  * `context`: provides utilities from the runtime environment, and is described in more detail there.
```

language/resource-api/README.md
```diff
@@ -208,18 +209,18 @@ class Puppet::Provider::AptKey::AptKey < Puppet::ResourceApi::SimpleProvider
...
+  * `name`: the name (if there is only one namevar) or hash of namevars (if there are multiple) of the new resource.
   * `should`: a hash of the attributes for the new instance.
 
 * `update(context, name, should)`: This is called when a resource should be updated.
-  * `context`: provides utilties from the runtime environment, and is decribed in more detail there.
-  * `name`: the name or hash of the resource to change.
+  * `context`: provides provides utilities from the runtime environment, and is described in more detail there.
```
hlindberg:
```suggestion
  * `context`: provides utilities from the runtime environment, and is described in more detail there.
```

language/resource-api/README.md
```diff
@@ -208,18 +209,18 @@ class Puppet::Provider::AptKey::AptKey < Puppet::ResourceApi::SimpleProvider
...
+  * `name`: the name (if there is only one namevar) or hash of namevars (if there are multiple) of the new resource.
   * `should`: a hash of the desired state of the attributes.
 
 * `delete(context, name)`: This is called when a resource should be deleted.
-  * `context`: provides utilties from the runtime environment, and is decribed in more detail there.
-  * `name`: the name or hash of the resource that should be deleted.
+  * `context`: provides provides utilities from the runtime environment, and is described in more detail there.
```
hlindberg:
```suggestion
  * `context`: provides utilities from the runtime environment, and is described in more detail there.
```

language/resource-api/README.md
```diff
@@ -143,6 +142,8 @@ software { php:
 }
 ```
 
+The provider should return a string containing all namevar values in a string matching one of the title patterns. This allows the runtime environment to present resources in a way familiar to the user.
```
clairecadman:
```suggestion
The provider returns a string containing all namevar values in a string matching one of the title patterns.
```

language/resource-api/README.md
```diff
@@ -184,7 +185,7 @@ A missing `:should` entry indicates that a resource should be removed from the s
 
 The `set` method should always return `nil`. Any progress signaling should be done through the logging utilities described below. If the `set` method throws an exception, all resources that should change in this call and haven't already been marked with a definite state, will be marked as failed. The runtime will only call the `set` method if there are changes to be made, especially in the case of resources marked with `noop => true` (either locally or through a global flag). The runtime will not pass them to `set`. See `supports_noop` below for changing this behaviour if required.
 
-Both methods take a `context` parameter which provides utilties from the runtime environment, and is decribed in more detail there.
+Both methods take a `context` parameter which provides utilities from the runtime environment, and is described in more detail there.
```
clairecadman: Where is this described? In the runtime environment?

language/resource-api/README.md
clairecadman: I don't think the last line adds anything, and avoid saying "the user", as they are the ones reading it.

hlindberg: Actually, now that this sentence was called out and I read it again, it does not really make sense - "returning a string containing all namevar values in a string" ... what does that actually mean. And the original said "should return", as in a requirement on the author of a provider; unclear if this is something an implementor of a resource type can do anything about...

Can meaning be made a more clear?

da-ar: > I don't think the last line adds anything, and avoid saying "the user", as they are the ones reading it.

For clarity, this documentation is directed at a module author, they must ensure that the provider that they are writing returns a `namevar` string that matches a `title-pattern` in a way that  "the user" (the eventual end user of the new module) understands.

hlindberg:
```suggestion
The provider should return a single formatted string representing all namevar values in such a way that it matches one of the title patterns. This allows the runtime environment to present resources in a way familiar to the user.
```
Now I get it - how about this suggestion?

DavidS:
I'll come back to this after contributor summit.

language/resource-api/README.md
DavidS: :+1:

language/resource-api/README.md
DavidS: the power of copy and paste :-1:

language/resource-api/README.md
DavidS: I've added the where and how bit, too.

language/resource-api/README.md
DavidS: yeah. changed it (and the three other occurrences) to "described in more detail in its own section below"

DavidS:
I've found a few more tickets that fold into the composite namevars issue that I'd like to bundle into this fix:
* https://tickets.puppetlabs.com/browse/PDK-1219 - purging multi-namevar resources fails when the provider does not return a title
* https://tickets.puppetlabs.com/browse/PDK-1220 - document that multi-namevar types should return a title from get
* https://tickets.puppetlabs.com/browse/PDK-1221 - add consistency checks for title attribute on multi-namevar providers

PDK-1220 is already covered in this text, the others need a bit more consideration.

----

against my usual preferences in this repo, I've rebased and munged the commits on this change as far as it went to separate the editorial and the content changes.

DavidS:
I've additional changes in flight at https://github.com/DavidS/puppet-specifications/compare/clarify-composite-namevars...DavidS:clarify-composite-namevars-tmp?expand=1 but leaving it off the main PR until I have everything together and the commits sorted out by purpose.

DavidS:
And rebased everything so that the diff linked above is sane.

language/resource-api/README.md
DavidS: I've rewritten this section extensively

DavidS:
@Lavinia-Dan @da-ar this is now in a shape to review. Note that the commits are broken up into distinct chunks that should help reduce clutter during reviewing.

language/resource-api/README.md
```diff
@@ -184,6 +184,24 @@ The `get` method reports the current state of the managed resources. It returns
...
+>   'name' => {
+>     is: {
+>       name: 'name',
+>       ensure: 'present',
+>       value: 'original value,
+>     },
+>     should: {
+>       name: 'name',
+>       ensure: 'present',
+>       value: 'new value,
```
da-ar: missing  end quote

language/resource-api/README.md
```diff
@@ -184,6 +184,24 @@ The `get` method reports the current state of the managed resources. It returns
...
 The `set` method updates resources to a new state. The `changes` parameter gets passed a hash of change requests, keyed by the resource's name. Each value is another hash with the optional `:is` and `:should` keys. At least one of the two has to be specified. The values will be of the same shape as those returned by `get`. After the `set`, all resources should be in the state defined by the `:should` values.
 
+> Example for a `changes` request:
+> ```
+> changes = {
+>   'name' => {
+>     is: {
+>       name: 'name',
+>       ensure: 'present',
+>       value: 'original value,
```
da-ar: missing  end quote

language/resource-api/README.md
```diff
@@ -230,18 +230,14 @@ class Puppet::Provider::AptKey::AptKey < Puppet::ResourceApi::SimpleProvider
...
-  * `should`: a hash of the desired state of the attributes.
-
 * `delete(context, name)`: This is called when a resource should be deleted.
-  * `context`: provides utilities from the runtime environment, and is described in more detail in its own section below.
-  * `name`: the name or hash of the resource that should be deleted.
+
+The parameters of these methods always carry the same values:
+
+* `context`: provides utilities from the runtime environment, and is described in more detail in its own section below.
+* `name`: the name (if there is only one namevar) or hash of namevars (if there are multiple) of the new resource.
```
da-ar: remove "new" at the end of this line. As should now read for update and delete

language/resource-api/README.md
```diff
@@ -214,15 +250,15 @@ In many cases, the resource type follows the conventional patterns of puppet, an
 
 `SimpleProvider` requires that your type follows some common conventions:
 
-* `name` is the name of your namevar attribute
+* the single namevar attribute is called `name` or the type has more than one namevar
```
da-ar: This doesn't make sense to me as a sentence. Perhaps:

A single namevar attribute called `name`; or, `names` if the type has multiple namevar attributes, which will receive a hash.

language/resource-api/README.md
```diff
@@ -69,16 +70,16 @@ The `Puppet::ResourceApi.register_type(options)` function takes the following ke
...
 
-Each resource being managed must be identified by a unique title. Usually this is fairly straightforward and a single attribute can be used to act as an identifier. Sometimes though, you need a composite of two attributes to uniquely identify the resource you want to manage.
+Each resource being managed must be identified by a unique title. Usually this is fairly straightforward and a single attribute can be used to act as an identifier. Sometimes though, you need a composite of two or more attributes to uniquely identify the resource you want to manage.
 
-If multiple attributes are defined with the `namevar` behaviour, the type SHOULD specify `title_patterns` that will tell Resource API how to get at the attributes from the title. If `title_patterns` is not specified a default pattern is applied, and matches against the first declared `namevar`.
+If multiple attributes are defined with the `namevar` behaviour, the type SHOULD specify `title_patterns` that will tell Resource API how to get at the attributes from the title. If `title_patterns` is not specified a default pattern is applied, and matches the entire title against the first declared `namevar`.
```
hlindberg: Is this `SHOULD` really a `MUST` - if not, how would it otherwise deal with it?

language/resource-api/README.md
```diff
@@ -69,16 +70,16 @@ The `Puppet::ResourceApi.register_type(options)` function takes the following ke
...
+> Note: Not supplying a title_pattern that assigns values to all namevars will severely restrict the usability of the provider.
```
hlindberg: So what should you instead do?

language/resource-api/README.md
```diff
@@ -265,7 +318,21 @@ The runtime environment needs to compare user input from the manifest (the desir
 
 The `canonicalize` method transforms its `resources` argument into the standard format required by the rest of the provider. The `resources` argument to `canonicalize` is an enumerable of resource hashes matching the structure returned by `get`. It returns all passed values in the same structure with the required transformations applied. It is free to reuse or recreate the data structures passed in as arguments. The runtime environment must use `canonicalize` before comparing user input values with values returned from `get`. The runtime environment always passes canonicalized values into `set`. If the runtime environment requires the original values for later processing, it protects itself from modifications to the objects passed into `canonicalize`, for example through creating a deep copy of the objects.
 
-The `context` parameter is the same passed to `get` and `set`, which provides utilties from the runtime environment, and is decribed in more detail there.
+The `context` parameter is the same passed to `get` and `set`, which provides utilities from the runtime environment, and is described in more detail there.
```
hlindberg: This could be rephrased for better readability. It is not `get` and `set` that "provides utilities"

language/resource-api/README.md
```diff
@@ -304,17 +371,22 @@ Puppet::ResourceApi.register_type(
...
+      return all_instances
+    else
+      names.collect { |n| find_instance(n) }
+    end
   end
 ```
 
 Some resources are very expensive to enumerate. The provider can implement `simple_get_filter` to signal extended capabilities of the `get` method to address this. The provider's `get` method will be called with an array of resource names, or `nil`. The `get` method must at least return the resources mentioned in the `names` array, but may return more than those. If the `names` parameter is `nil`, all existing resources should be returned. The `names` parameter defaults to `nil` to allow simple runtimes to ignore this feature.
 
+For types with multiple namevars, the `names` array will consist of hashes of the namevars and their variables instead of simple values:
```
hlindberg: This sounds odd. "namevars and their variables"...

language/resource-api/README.md
DavidS: :+1:

language/resource-api/README.md
DavidS: :+1:

language/resource-api/README.md
DavidS: :+1:

language/resource-api/README.md
DavidS: changed it back to "if the type has a single namevar, it must be called `name`"

looking at the code, I *think* we can relax that, too. but that's a conversation for another day.

language/resource-api/README.md
DavidS: As the text says: if `title_patterns` is not specified a default pattern is applied, and matches the entire title against the first declared `namevar`. This will mean that in puppet code all other namevars need to be specified as attributes in the body of the resource.

language/resource-api/README.md
DavidS: see below/

language/resource-api/README.md
DavidS: :+1:

language/resource-api/README.md
DavidS: yup. should have been "namevars and their values". fixed.

language/resource-api/README.md
da-ar: 👍

language/resource-api/README.md
hlindberg: I still feel line 80 is dangling - comes across as a threat as I don't really understand what it is telling me. I understand line 79, it is telling me what I should/must do.

hlindberg: Ah... I think it is saying "A supplied title pattern MUST assign values to all namevars, as not doing this will severely restrict the usability of the provider."
Did I get it?

hlindberg: Sorry to harp on this, but since I don't know how it actually works and would like to know... I keep coming back to this :-)
When you say "As the text says: if title_patterns is not specified a default pattern is applied, and matches the entire title against the first declared namevar. This will mean that in puppet code all other namevars need to be specified as attributes in the body of the resource." - is that in any way recommended way of doing it? Are people doing it? If so, you may want to include something about this - "In legacy code this may have been implemented as <some variant of what you wrote> ... This is discouraged (deprecated even?) as it strongly restricts the usability of the provider"

DavidS: I'm trying to toe the line between "it's not required" and "if you try and it breaks, you may keep both pieces".

I should just stamp down on it and make not having a `title_pattern` that assigns all namevars an error.

hlindberg: Yes, and then change the `SHOULD` to `MUST` - we take no responsibility, nor do we tell you what will happen - it is simply undefined and W.R.O.N.G.

DavidS: Yeah, I'll do that - even if it's possibly not something we can actively check on the backend. :man_shrugging:

DavidS:
Ok, I think this wraps it up. The implementation is available at https://github.com/puppetlabs/puppet-resource_api/pull/174 and there don't seem to be further comments here.

language/resource-api/README.md
```diff
@@ -176,58 +177,109 @@ class Puppet::Provider::AptKey::AptKey
...
+  >     ensure: 'present',
+  > }
+  > ```
 
 The `set` method updates resources to a new state. The `changes` parameter gets passed a hash of change requests, keyed by the resource's name. Each value is another hash with the optional `:is` and `:should` keys. At least one of the two has to be specified. The values will be of the same shape as those returned by `get`. After the `set`, all resources should be in the state defined by the `:should` values.
 
+> Example for a `changes` request:
+> ```
+> changes = {
+>   'name' => {
```
da-ar: I think this example would be clearer if the name value was something like 'foo' . 
```
changes = {
  'foo' => {
    is: {
      name: 'foo',
      ensure: 'present',
      value: 'original value',
    },
    should: {
      name: 'foo',
      ensure: 'present',
      value: 'new value',
    },
  },
}
```

PR #141, opened 2019-05-29
(maint) fix desc/docs confusion

DavidS:
The text and examples have been inconsistent with how `desc` vs `docs`
has been handled. This change fixes the text and examples to all show
and require `desc`, but accept `docs` in places where we showed it
previously.

### DIFF
--- a/language/resource-api/README.md
+++ b/language/resource-api/README.md
@@ -56,7 +56,7 @@ Puppet::ResourceApi.register_type(
 The `Puppet::ResourceApi.register_type(options)` function takes the following keyword arguments:
 
 * `name`: the name of the resource type.
-* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues.
+* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues. Can also be written as `docs` for backwards compatibility. This compatibility option will go away in the next major revision of the library.
 * `attributes`: a hash mapping attribute names to their details. Each attribute is described by a hash containing the Puppet 4 data `type`, a `desc` string, a `default` value, and the `behaviour` of the attribute: `namevar`, `read_only`, `init_only`, or a `parameter`.
   * `type`: the Puppet 4 data type allowed in this attribute.
   * `desc`: a string describing this attribute. This is used in creating the automated API docs with [puppet-strings](https://github.com/puppetlabs/puppet-strings).
@@ -90,7 +90,7 @@ Example:
 # lib/puppet/type/software.rb
 Puppet::ResourceApi.register_type(
   name: 'software',
-  docs: <<-DOC,
+  desc: <<-DOC,
     This type provides Puppet with the capabilities to manage ...
   DOC
   title_patterns: [
--- a/language/resource-api/README.md
+++ b/language/resource-api/README.md
@@ -56,7 +56,7 @@ Puppet::ResourceApi.register_type(
 The `Puppet::ResourceApi.register_type(options)` function takes the following keyword arguments:
 
 * `name`: the name of the resource type.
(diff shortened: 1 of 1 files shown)


language/resource-api/README.md
```diff
@@ -56,7 +56,7 @@ Puppet::ResourceApi.register_type(
 The `Puppet::ResourceApi.register_type(options)` function takes the following keyword arguments:
 
 * `name`: the name of the resource type.
-* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues.
+* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues. Can also be written as `docs` for backwards compatibility. This compatibility option will go away in the next major revision.
```
hlindberg: The term "next major revision" does not work so well since it is a reference to the resource-api next major version as opposed to a revision of the specification. Maybe say "in a future major revision"?

hlindberg: hm.. or is this meant to be a "revision of the spec"  - in which case we should perhaps formalize the versions of the specifications - since the entire repo covers "all the things"...

DavidS: good point. I'll clarify to mean the Resource API gem.

raises the wider question of whether this is still a good place to keep this document or if it were better to put this spec into the https://github.com/puppetlabs/puppet-resource_api repo and synchronize it there with the different release train branches.

language/resource-api/README.md
```diff
@@ -56,7 +56,7 @@ Puppet::ResourceApi.register_type(
 The `Puppet::ResourceApi.register_type(options)` function takes the following keyword arguments:
 
 * `name`: the name of the resource type.
-* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues.
+* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues. Can also be written as `docs` for backwards compatibility. This compatibility option will go away in the next major revision of the library.
```
clairecadman:
```suggestion
* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues. You can also be write `desc` as `docs` for backwards compatibility. This compatibility option will no longer be available in the next major revision of the library.
```

language/resource-api/README.md
hlindberg: It makes sense for it to be here if there can be alternative implementations of the API, but must then be written to be free of implementation concerns. For example, this could be written as "an implementation may support the older `docs` tag to have the same meaning as `desc`. And add a note that PDK version such and such supports, and is expected to drop in...

This could be pedantic and somewhat silly when there is only one project implementing the specification.
Just my 2c.

DavidS: There is only one implementation and any compatible other implementation would need to implement this fallback too to be compatible. I'll leave it as it is for now.

language/resource-api/README.md
```diff
@@ -56,7 +56,7 @@ Puppet::ResourceApi.register_type(
 The `Puppet::ResourceApi.register_type(options)` function takes the following keyword arguments:
 
 * `name`: the name of the resource type.
-* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues.
+* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues. You can also write `desc` as `docs` for backwards compatibility. This compatibility option will no longer be available in the next major revision of the library.
```
hlindberg: Will you change this text once the "next major revision of the library" is released?

DavidS: I think I've been pretty good until now to keep this text and the library in sync. I expect to do so in the foreseeable future, too.

DavidS:
Merging this as the implementation is now merged.

PR #142, opened 2019-06-12
(PUP-9747) update bolt transport key description

DavidS:
* Emphasise source of design constraint
* Emphasise recommendation to use these keys when applicable
* Add `password` key following feedback from the bolt team

### DIFF
--- a/language/resource-api/README.md
+++ b/language/resource-api/README.md
@@ -432,23 +432,26 @@ Password attributes should also set `sensitive: true` to ensure that the data is
 
 #### Schema
 
-A transport schema accepts the following keywords:
+To line up with [bolt's inventory file](https://puppet.com/docs/bolt/latest/inventory_file.html), a transport schema should prefer the following keywords when relevant:
 
-* `uri`: use when you need to specify a specific URL to connect to. All of the following keys will be computed from the `uri` when possible. In the future more url parts may be computed from the URI.
-* `host`: use to specify an IP or address to connect to.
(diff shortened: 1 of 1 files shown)


language/resource-api/README.md
```diff
@@ -432,23 +432,26 @@ Password attributes should also set `sensitive: true` to ensure that the data is
...
 
+* `implementations`: reserved by Bolt.
 * `name`: transports should use `uri` instead of name.
 * `path`: reserved as a uri part.
 * `query`: reserved as a uri part.
-* `run-on`: This is used by Bolt to determine which target to proxy to. Transports should not rely on this key.
-* `remote-transport`: This is used to determine which transport to load. It should always be the transport class named "declassified".
 * `remote-*`: any key starting with `remote-` is reserved for future use.
-* `implementations`: reserved by Bolt.
+* `remote-transport`: determines which transport to load. It is always the transport class named "declassified".
```
da-ar: > It is always the transport class named "declassified".

The sentence doesn't make sense to me.  It reads like `remote-transport: declassified`. 

Should it be: It is always the transport class name "declassified".` 

If so, I believe it may make more sense to be written as:
It is always the transport class name written in lowercase. e.g. `class Panos` becomes `panos`

DavidS: It should always be `the transport name`. The class name for the transport is the snake case (`SnakeCase`) version of the transport name.

good spot, @da-ar

PR #145, opened 2019-11-08
Allow task parameters to have default values

nicklewis:
This adds a "default" key to the parameter specification to set a value 
that should be passed to the task if the parameter isn't provided. This 
value must satisfy the "type" of the parameter.

This should be backward-compatible so long as the task implementation is
correctly written to match the parameter type specification. Specifically, task
authors should understand that the "default" field is advisory to the task
runner and should not assume that a parameter of type `Optional` will always
have a value just because it has a "default" specified.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -29,12 +29,13 @@ These are the revisions to this version of the task spec.
 
 **Note**: Bolt generally uses the latest version and revision of the task spec. Puppet Enterprise versions use a specific version and revision of the task spec, indicated in the PE documentation.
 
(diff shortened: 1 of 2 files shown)


nicklewis:
Updated to mention the task runner must respect default values

tasks/README.md
```diff
@@ -269,6 +277,8 @@ Future releases of the task runner may expand this or add configuration options
 
 ## Input Methods
```
adreyer:
```suggestion
## Input
```

PR #151, opened 2021-02-04
Fix use of unknown method .error

vStone:
The resource_api implements `context.err` but not `context.error` as documented here.


DavidS (approved):
This one ALWAYS gets me, after over a decade, every time :-(

PR #153, opened 2021-04-19
(GH-225) Add custom_insync to rsapi docs

michaeltlombardi:
This PR coincides with the work for puppetlabs/puppet-resource_api#225
to add the new custom_insync feature and functionality to rsapi resources;

It covers a common use case in the code and explains the purpose and some
implementation concerns.

### DIFF
--- a/language/resource-api/README.md
+++ b/language/resource-api/README.md
@@ -2,7 +2,7 @@
 
 This library provides a simple way to write new native resources for [Puppet](https://puppet.com).
 
-A *resource* is the basic unit that is managed by Puppet. Each resource has a set of attributes describing its current state. Some attributes can be changed throughout the lifetime of the resource, whereas others are only reported back but cannot be changed (see `read_only`), and some can only be set once during initial creation (see `init_only`). To gather information about those resources and to enact changes, Puppet requires a *provider* to implement this interaction. The provider can have parameters that influence its mode of operation (see `parameter`). To describe all these parts to the infrastructure and the consumers, the resource *type* defines all the metadata, including the list of the attributes. The *provider* contains the code to *get* and *set* the system state.
+A _resource_ is the basic unit that is managed by Puppet. Each resource has a set of attributes describing its current state. Some attributes can be changed throughout the lifetime of the resource, whereas others are only reported back but cannot be changed (see `read_only`), and some can only be set once during initial creation (see `init_only`). To gather information about those resources and to enact changes, Puppet requires a _provider_ to implement this interaction. The provider can have parameters that influence its mode of operation (see `parameter`). To describe all these parts to the infrastructure and the consumers, the resource _type_ defines all the metadata, including the list of the attributes. The _provider_ contains the code to _get_ and _set_ the system state.
 
 ## Resource definition ("type")
 
@@ -56,23 +56,23 @@ Puppet::ResourceApi.register_type(
 
 The `Puppet::ResourceApi.register_type(options)` function takes the following keyword arguments:
 
-* `name`: the name of the resource type.
-* `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues. You can also write `desc` as `docs` for backwards compatibility. This compatibility option will no longer be available in the next major revision of the library.
-* `attributes`: a hash mapping attribute names to their details. Each attribute is described by a hash containing the Puppet 4 data `type`, a `desc` string, a `default` value, and the `behaviour` of the attribute: `namevar`, `read_only`, `init_only`, or a `parameter`.
-  * `type`: the Puppet 4 data type allowed in this attribute.
-  * `desc`: a string describing this attribute. This is used in creating the automated API docs with [puppet-strings](https://github.com/puppetlabs/puppet-strings).
-  * `default`: a default value that will be used by the runtime environment when the caller does not specify a value for this attribute.
-  * `behaviour`/`behavior`: how the attribute behaves. Currently available values:
-    * `namevar`: marks an attribute as part of the "primary key" or "identity" of the resource. A given set of `namevar` values needs to distinctively identify an instance.
-    * `init_only`: this attribute can only be set during the creation of the resource. Its value will be reported going forward, but trying to change it later will lead to an error. For example, the base image for a VM or the UID of a user.
-    * `read_only`: values for this attribute will be returned by `get()`, but `set()` is not able to change them. Values for this should never be specified in a manifest. For example, the checksum of a file or the MAC address of a network interface.
-    * `parameter`: these attributes influence how the provider behaves and cannot be read from the target system. For example, the target file on inifile or credentials to access an API.
-* `autorequire`, `autobefore`, `autosubscribe`, and `autonotify`: a hash mapping resource types to titles. The titles must either be constants, or, if the value starts with a dollar sign, a reference to the value of an attribute. If the specified resources exist in the catalog, Puppet will create the relationsships requested here.
-* `features`: a list of API feature names, specifying which optional parts of this spec the provider supports. Currently defined features: `canonicalize`, `simple_get_filter`, and `supports_noop`. See below for details.
+- `name`: the name of the resource type.
+- `desc`: a doc string that describes the overall working of the resource type, provides examples, and explains prerequisites and known issues. You can also write `desc` as `docs` for backwards compatibility. This compatibility option will no longer be available in the next major revision of the library.
+- `attributes`: a hash mapping attribute names to their details. Each attribute is described by a hash containing the Puppet 4 data `type`, a `desc` string, a `default` value, and the `behaviour` of the attribute: `namevar`, `read_only`, `init_only`, or a `parameter`.
+  - `type`: the Puppet 4 data type allowed in this attribute.
+  - `desc`: a string describing this attribute. This is used in creating the automated API docs with [puppet-strings](https://github.com/puppetlabs/puppet-strings).
+  - `default`: a default value that will be used by the runtime environment when the caller does not specify a value for this attribute.
+  - `behaviour`/`behavior`: how the attribute behaves. Currently available values:
+    - `namevar`: marks an attribute as part of the "primary key" or "identity" of the resource. A given set of `namevar` values needs to distinctively identify an instance.
+    - `init_only`: this attribute can only be set during the creation of the resource. Its value will be reported going forward, but trying to change it later will lead to an error. For example, the base image for a VM or the UID of a user.
+    - `read_only`: values for this attribute will be returned by `get()`, but `set()` is not able to change them. Values for this should never be specified in a manifest. For example, the checksum of a file or the MAC address of a network interface.
+    - `parameter`: these attributes influence how the provider behaves and cannot be read from the target system. For example, the target file on inifile or credentials to access an API.
+- `autorequire`, `autobefore`, `autosubscribe`, and `autonotify`: a hash mapping resource types to titles. The titles must either be constants, or, if the value starts with a dollar sign, a reference to the value of an attribute. If the specified resources exist in the catalog, Puppet will create the relationsships requested here.
+- `features`: a list of API feature names, specifying which optional parts of this spec the provider supports. Currently defined features: `canonicalize`, `custom_insync`, `simple_get_filter`, and `supports_noop`. See below for details.
 
 For autoloading to work, this code needs to go into `lib/puppet/type/<name>.rb` in your module.
 
-###  Composite Namevars ("title_patterns")
+### Composite Namevars ("title_patterns")
 
 Each resource being managed must be identified by a unique title. Usually this is fairly straightforward and a single attribute can be used to act as an identifier. Sometimes though, you need a composite of two or more attributes to uniquely identify the resource you want to manage.
 
@@ -81,8 +81,9 @@ If multiple attributes are defined with the `namevar` behaviour, the type must s
 The `title_patterns` are evaluated in the order they are specified. Evaluation stops after the first match is achieved.
 
 Each title pattern contains the:
-  * `pattern`, which is a ruby regex containing named captures. The names of the captures MUST be that of the namevar attributes.
-  * `desc`, a short description of what the pattern matches for.
+
+- `pattern`, which is a ruby regex containing named captures. The names of the captures MUST be that of the namevar attributes.
+- `desc`, a short description of what the pattern matches for.
 
 Example:
 
@@ -124,6 +125,7 @@ Puppet::ResourceApi.register_type(
 ```
 
 Matches the first title pattern:
+
 ```puppet
 # /etc/puppetlabs/code/environments/production/manifests/site.pp
 software { php-yum:
@@ -136,6 +138,7 @@ software { php-gem:
 ```
 
 Matches the second title pattern:
+
 ```puppet
 # /etc/puppetlabs/code/environments/production/manifests/site.pp
 software { php:
@@ -179,13 +182,15 @@ end
 
 The `get` method reports the current state of the managed resources. It returns an enumerable of all existing resources. Each resource is a hash with attribute names as keys, and their respective values as values.
 
-* It is an error to return values not matching the type specified in the resource type.
-* If a requested resource is not listed in the result, it is considered to not exist on the system.
-* If the `get` method raises an exception, the provider is marked as unavailable during the current run, and all resources of this type will fail in the current transaction. The exception message will be reported to the user.
-* If the type has more than one namevar, each resource hash must have a `:title` key with a single formatted string representing all namevar values. The title must match one of the title patterns. The namevar values computed from that pattern must match their counterparts in the hash.
+- It is an error to return values not matching the type specified in the resource type.
+- If a requested resource is not listed in the result, it is considered to not exist on the system.
+- If the `get` method raises an exception, the provider is marked as unavailable during the current run, and all resources of this type will fail in the current transaction. The exception message will be reported to the user.
+- If the type has more than one namevar, each resource hash must have a `:title` key with a single formatted string representing all namevar values. The title must match one of the title patterns. The namevar values computed from that pattern must match their counterparts in the hash.
+
   > Note: This value allows the runtime environment to present resources in a way familiar to the user.
 
   > Example: Referring back to the `software` example above, resources returned from the `get` method require a `title:` value matching the `package:` and `manager:` values:
+  >
   > ```
   > {
   >     title: 'php-yum',
@@ -198,6 +203,7 @@ The `get` method reports the current state of the managed resources. It returns 
 The `set` method updates resources to a new state. The `changes` parameter gets passed a hash of change requests, keyed by the resource's name. Each value is another hash with the optional `:is` and `:should` keys. At least one of the two has to be specified. The values will be of the same shape as those returned by `get`. After the `set`, all resources should be in the state defined by the `:should` values.
 
 > Example for a `changes` request:
+>
 > ```
 > changes = {
 >   'foo' => {
@@ -220,6 +226,7 @@ A missing `:should` entry indicates that a resource should be removed from the s
 If a type has more than one namevar, the resource's name key is replaced by a hash of the namevars and their values:
 
 > Example for a `changes` request with multiple namevars:
+>
 > ```
 > changes = {
 >   { package: 'php', manager: 'yum' } => {
@@ -249,8 +256,8 @@ In many cases, the resource type follows the conventional patterns of puppet, an
 
 `SimpleProvider` requires that your type follows some common conventions:
 
-* if the type has a single namevar attribute, it must be called `name`
-* `ensure` attribute is present and has the `Enum[absent, present]` type
+- if the type has a single namevar attribute, it must be called `name`
+- `ensure` attribute is present and has the `Enum[absent, present]` type
 
 To start using `SimpleProvider`, inherit from the class like this:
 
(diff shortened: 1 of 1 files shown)


language/resource-api/README.md
```diff
@@ -356,6 +356,43 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+  name: 'apt_key',
+  features: [ 'custom_insync' ],
+)
+
+# lib/puppet/provider/apt_key/apt_key.rb
+class Puppet::Provider::AptKey::AptKey
+  def insync?(context, name, attribute_name, is_hash, should_hash)
+    case attribute_name
+    when :some_array_attribute
+      if should_hash[:force]
```
DavidS, line 375: please do define the attributes used here in the type, so that it is clear how they are defined (property vs parameter).

Since this is also only called for properties, it's `property_name`, not `attribute_name` in the sig. Also needs to be changed in the test in https://github.com/puppetlabs/puppet-resource_api/pull/285

language/resource-api/README.md
```diff
@@ -356,6 +356,43 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+        context.notice("Checking a subset match array")
+        # This will return true so long as all values in the array for
+        # the should_hash are also in the array of the is_hash.
+        return (is_hash[attribute_name] & should_hash[attribute_name]).sort == should_hash[attribute_name].sort
+      end
+    end
+  end
+```
+
+Specifying the `custom_insync` feature flag ensures that all of the property attributes on the type will call the `insync?` method in the provider, if it exists, enabling custom state comparisons as complex or simple as required. The `insync?` method in the provider has access to the `context` for logging, the name of the resource being checked, the attribute that is currently being inspected for insync status, and hashes representing the current state of the resource as well as the canonicalized should state of the resource.
```
DavidS, line 390:
```suggestion
When the `custom_insync` feature flag is specified, `insync?` is called once for each of the properties. This can be used to implement custom state comparisons, when the default value-equality is not sufficient. The `insync?` method receives a `context` for logging, the `name` of the resource being checked, the `property_name` that is currently being inspected for insync status, the current state of the resource in `is_hash`, as well as the desired state (canonicalized) in `should_hash`.
```

language/resource-api/README.md
```diff
@@ -356,6 +356,43 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+
+The `insync?` method is only and always triggered for each of the property attributes of the type.
```
DavidS, line 392:
```suggestion
The `insync?` method is executed exactly once for each of the properties of the type per resource instance.
The call happens during application of the catalog when puppet decides whether or not it needs to enforce state on a specific resource.
```

language/resource-api/README.md
```diff
@@ -356,6 +356,43 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+
+If the custom `insync?` method in the provider must return a truthy or falsey result, or `nil`; if the return is `nil` the comparison defaults to the existing comparison logic for properties in Puppet. This means that unless the `insync?` method in the provider has special handling for the attribute Puppet will default to the existing state comparison behaviors. Therefore, only properties which need special handling need to be addressed in the `insync?` method in the provider.
```
DavidS, line 394:
```suggestion
The `insync?` method's return value will be interpreted as truthy or falsey, following the regular ruby rules, except `nil` values; if the return is `nil` the comparison defaults to the existing comparison logic for properties in Puppet. This means that unless the `insync?` method in the provider has special handling for the attribute Puppet will default to the existing state comparison behaviors. Therefore, only properties which need special handling need to be addressed in the `insync?` method in the provider.
```

DavidS (changes requested):

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+    end
+end
+```
+
+When the `custom_insync` feature flag is specified, `insync?` is called once for each of the properties.
+This can be used to implement custom state comparisons, when the default value-equality is not sufficient.
+The `insync?` method receives a `context` for logging, the `name` of the resource being checked, the `property_name` that is currently being inspected for insync status, the current state of the resource in `is_hash`, as well as the desired state (canonicalized) in `should_hash`.
+
+The `insync?` method is executed exactly once for each of the properties of the type per resource instance.
+The call happens during application of the catalog when puppet decides whether or not it needs to enforce state on a specific resource.
```
DavidS, line 419:
```suggestion
The call happens during application of the catalog when the runtime environment decides whether or not it needs to enforce state on a specific resource.
```

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+
+The `insync?` method's return value will drive the behavior of Puppet for that property:
+
+- If `insync?` returns `nil`, property comparison is handed back to Puppet's built-in property insync checking.
```
DavidS, line 423:
```suggestion
- If `insync?` returns `nil`, property comparison is handed back to the runtime environment's built-in property value equality checking.
```

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+  - This means that _only_ properties which actually need custom insync logic need to be handled in the `insync?` method on the provider. Everything else can be ignored, which will return `nil` and use default comparison logic.
```
DavidS, line 424:
```suggestion
  - This means that the provider _only_ needs to deal with properties which actually need custom insync logic. All other properties can be ignored, which will return ruby's default `nil` return value and use default comparison logic.
```

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+- If `insync?` returns `true`, the property is treated as insync.
+- If `insync?` returns `false`, the property is treated as out of sync and will surface the default Puppet change reporting.
```
DavidS, line 426:
```suggestion
- If `insync?` returns `false`, the property is treated as out of sync and will cause the default change reporting.
```

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+  - e.g. `changed <is value> to <should value>`
+- If `insync?` returns a string, the property is treated as out of sync and the string is surfaced in the Puppet change reporting.
```
DavidS, line 428:
```suggestion
- If `insync?` returns a string, the property is treated as out of sync and the string is used for change reporting instead of the default. This can be used to provide additional information about the reason for a change.
```

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+The `insync?` method is executed exactly once for each of the properties of the type per resource instance.
+The call happens during application of the catalog when puppet decides whether or not it needs to enforce state on a specific resource.
+
+The `insync?` method's return value will drive the behavior of Puppet for that property:
+
+- If `insync?` returns `nil`, property comparison is handed back to Puppet's built-in property insync checking.
+  - This means that _only_ properties which actually need custom insync logic need to be handled in the `insync?` method on the provider. Everything else can be ignored, which will return `nil` and use default comparison logic.
+- If `insync?` returns `true`, the property is treated as insync.
+- If `insync?` returns `false`, the property is treated as out of sync and will surface the default Puppet change reporting.
+  - e.g. `changed <is value> to <should value>`
```
DavidS, line 427:
```suggestion
  - e.g. `Notice: /Stage[main]/Main/File[/tmp/foo]/mode: mode changed <is value> to <should value>`
```

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+- If `insync?` returns a string, the property is treated as out of sync and the string is surfaced in the Puppet change reporting.
+  - In the `test_custom_insync` provider example above, you would get the message `Adding missing members ['SomeValue']` if `SomeValue` was specified as a member of `some_array` but not found in the actual state returned from `get`.
```
DavidS, line 429:
```suggestion
  - In the `test_custom_insync` provider example above, you would get the message `Notice: /Stage[main]/Main/Test_custom_insync[something]/some_array: Adding missing members ['SomeValue']` if `SomeValue` was specified as a member of `some_array` but not found in the actual state returned from `get`.
```

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+    changes.each do |_name, is_and_should|
+      context.notice("Setting with #{is_and_should[:should].inspect}")
+    end
+  end
+
+  def insync?(context, name, _property_name, is_hash, should_hash)
+    context.notice("Checking whether #{name} is out of sync")
+
+    return true unless should_hash[:force]
+    context.notice("Out of sync!")
```
DavidS, line 483:
```suggestion
    context.debug("Checking whether #{name} is out of sync")

    return true unless should_hash[:force]
    context.debug("Out of sync!")
```

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+        missing_members = should_hash[property_name].reject { |member| found_members.include?(member) }
+        return missing_members.empty? ? true : "Adding missing members #{missing_members}"
+      end
+    end
+end
+```
+
+When the `custom_insync` feature flag is specified, `insync?` is called once for each of the properties.
+This can be used to implement custom state comparisons, when the default value-equality is not sufficient.
+The `insync?` method receives a `context` for logging, the `name` of the resource being checked, the `property_name` that is currently being inspected for insync status, the current state of the resource in `is_hash`, as well as the desired state (canonicalized) in `should_hash`.
```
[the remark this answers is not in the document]
michaeltlombardi: Moved to after the return value explanation.

language/resource-api/README.md
```diff
@@ -356,6 +356,136 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+
+The `insync?` method is executed exactly once for each of the properties of the type per resource instance.
```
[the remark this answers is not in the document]
michaeltlombardi: Modified to read:

> The `insync?` method is executed exactly once for each of the **properties** (not parameters, read-only, or init attributes) of the type per resource instance.
> If no properties were specified in the instance of a resource, `insync?` is not called.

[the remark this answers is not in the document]
michaeltlombardi: wasn't sure on the wording for read-only/init.

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+      end
+    end
+end
+```
+
+In normal operations, the runtime environment uses strict value equality to determine if a resource instance needs to be state-enforced.
+When the `custom_insync` feature flag is specified, `insync?` is called once for each of the properties.
+This can be used to implement custom state comparisons, when the default value-equality is not sufficient.
+
+The `insync?` method is executed exactly once for each of the **properties** (not parameters, read-only, or init attributes) of the type per resource instance.
```
clairecadman, line 425:
```suggestion
The `insync?` method is executed once for each of the **properties** (not parameters, read-only, or init attributes) of the type per resource instance.
```

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+
+The `insync?(context, name, property_name, is_hash, should_hash)` method receives the following arguments:
+
+- a `context` to access the runtime environment,
+- the `name` of the resource being checked,
+- the `property_name` that is currently being inspected for insync status,
+- the current state of the resource in `is_hash`, and
+- the desired state (canonicalized) in `should_hash`.
+
+For resources which specify the `custom_insync` feature flag but have _no_ defined properties the Resource API adds a spectial hidden property, `rsapi_custom_insync_trigger`, which ensures that `insync?` is still called for the resource.
```
clairecadman, line 448:
```suggestion
For resources which specify the `custom_insync` feature flag but have _no_ defined properties, the Resource API adds a spectial hidden property —`rsapi_custom_insync_trigger` —  which ensures that `insync?` is still called for the resource.
```

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
 > values changed
 ```
 
+### Provider feature: `custom_insync`
+
+Allows for overriding default property `insync?` checking using custom logic in the provider.
```
clairecadman, line 368:
```suggestion
The `custom_insyc` feature lets you for override default property `insync?` checking using custom logic in the provider.
```

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+        context.debug("Checking a subset match array")
+        found_members = (is_hash[property_name] & should_hash[property_name]).sort
+        missing_members = should_hash[property_name].reject { |member| found_members.include?(member) }
+        return missing_members.empty? ? true : "Adding missing members #{missing_members}"
+      end
+    end
+end
+```
+
+In normal operations, the runtime environment uses strict value equality to determine if a resource instance needs to be state-enforced.
```
clairecadman, line 421: Does this mean without using this do feature?

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+When the `custom_insync` feature flag is specified, `insync?` is called once for each of the properties.
```
clairecadman, line 422:
```suggestion
When you specify the `custom_insync` feature flag, `insync?` is called once for each of the properties.
```

language/resource-api/README.md
michaeltlombardi: It does.

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+This can be used to implement custom state comparisons, when the default value-equality is not sufficient.
+
+The `insync?` method is executed exactly once for each of the **properties** (not parameters, read-only, or init attributes) of the type per resource instance.
+If no properties were specified in the instance of a resource, `insync?` is not called.
+The call happens during application of the catalog when the runtime environment decides whether or not it needs to enforce state on a specific resource.
+
+The `insync?` method's return value will drive the behavior of Puppet for that property:
```
clairecadman, line 429:
```suggestion
The `insync?` method's return value changes the way Puppet handles the property. For example:
```

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+- the current state of the resource in `is_hash`, and
+- the desired state (canonicalized) in `should_hash`.
+
+For resources which specify the `custom_insync` feature flag but have _no_ defined properties the Resource API adds a spectial hidden property, `rsapi_custom_insync_trigger`, which ensures that `insync?` is still called for the resource.
+This hidden property is not used or seen anywhere else in the provider, only in the `insync?` method.
+
+As with all calls to `insync?`, a custom change reporting message can be surfaced by returning a string instead of `true` or `false`.
+If no custom change reporting message is returned as a string, the change reporting in Puppet (if `false` is returned) will always be:
+
+> Custom insync logic determined that this resource is out of sync
```
clairecadman, line 454: Is this the message the user would see?

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+      type:      'Boolean',
+      desc:      'If true, the resource will be treated as not being insync.',
+      behaviour: :parameter,
+      default:   false,
+    },
+  },
+)
+
+# lib/puppet/provider/test_custom_insync_hidden_property/test_custom_insync_hidden_property.rb
+class Puppet::Provider::TestCustomInsyncHiddenProperty::TestCustomInsyncHiddenProperty
```
clairecadman, line 480: What is this example showing compared to the first one?

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+        missing_members = should_hash[property_name].reject { |member| found_members.include?(member) }
+        return missing_members.empty? ? true : "Adding missing members #{missing_members}"
+      end
+    end
+end
+```
+
+In normal operations, the runtime environment uses strict value equality to determine if a resource instance needs to be state-enforced.
+When the `custom_insync` feature flag is specified, `insync?` is called once for each of the properties.
+This can be used to implement custom state comparisons, when the default value-equality is not sufficient.
```
clairecadman, line 423:
```suggestion
You can use `custom_insync` to implement custom state comparisons when the default value-equality is not sufficient.
```

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+
+The `insync?` method is executed exactly once for each of the **properties** (not parameters, read-only, or init attributes) of the type per resource instance.
+If no properties were specified in the instance of a resource, `insync?` is not called.
+The call happens during application of the catalog when the runtime environment decides whether or not it needs to enforce state on a specific resource.
```
clairecadman, line 427: Is this desired state?

language/resource-api/README.md
michaeltlombardi: Yes.

language/resource-api/README.md
michaeltlombardi: This shows an example _without_ an insyncable property - which adds a hidden property to ensure that `insync?` is triggered.

language/resource-api/README.md
michaeltlombardi: It determines if the resource is _in_ the desired state.

language/resource-api/README.md
clairecadman: Maybe say "When you don't use "custom_insync, the runtime environment etc...."?

language/resource-api/README.md
clairecadman: Ah okay, maybe make that clearer?

language/resource-api/README.md
```diff
@@ -355,6 +363,144 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+You can use `custom_insync` to implement custom state comparisons when the default value-equality is not sufficient.
+
+The `insync?` method is executed once for each of the **properties** (not parameters, read-only, or init attributes) of the type per resource instance.
+If no properties were specified in the instance of a resource, `insync?` is not called.
+The call happens during application of the catalog when the runtime environment decides whether or not it needs to enforce state on a specific resource.
+
+The `insync?` method's return value changes the way Puppet handles the property. For example:
+
+- If `insync?` returns `nil`, property comparison is handed back to the runtime environment's built-in property value equality checking.
+  - This means that the provider _only_ needs to deal with properties which actually need custom insync logic. All other properties can be ignored, which will return ruby's default `nil` return value and use default comparison logic.
```
clairecadman, line 432: Capital R for Ruby.

language/resource-api/README.md
michaeltlombardi: I believe I've addressed this by breaking the docs up a little to clarify:

https://github.com/puppetlabs/puppet-specifications/pull/153/files#diff-23f362101839bf8ed43d75a53178cf3cb35aa2d8169fc0163f3a7619f87ef79fR421-R424

language/resource-api/README.md
michaeltlombardi: I've moved this entire section into a subheading, does that address this concern?

https://github.com/puppetlabs/puppet-specifications/pull/153/files#diff-23f362101839bf8ed43d75a53178cf3cb35aa2d8169fc0163f3a7619f87ef79fR457

michaeltlombardi:
@clairecadman just pinging to see if my updates addressed your review comments :)

language/resource-api/README.md
clairecadman: Maybe rephrase it like this?

language/resource-api/README.md
clairecadman: Yes that is better, thanks!

clairecadman:
@michaeltlombardi yes all good thanks! I don't want to change anything else as it would make it inconsistent with the rest of the README.

language/resource-api/README.md
```diff
@@ -355,6 +363,155 @@ For example, in the Puppet runtime environment this is bound to the `strict` set
...
+When you do not specify the `custom_insync` feature flag, the runtime environment uses strict value equality to determine if a resource instance needs to be state-enforced.
+
+When you **do** specify the `custom_insync` feature flag, `insync?` is called once for each of the properties.
+You can use `custom_insync` to implement custom state comparisons when the default value-equality is not sufficient.
+
+The `insync?` method is executed once for each of the **properties** (not parameters, read-only, or init attributes) of the type per resource instance.
+If no properties were specified in the instance of a resource, `insync?` is not called.
+The call happens during application of the catalog when the runtime environment decides whether or not it needs to enforce state on a specific resource.
+
+The `insync?` method **must** return `nil`, `true` or `false` as the result of the property comparison:
```
DavidS, line 430: this is not explaining the new API requested by @joshcooper

michaeltlombardi: I think this is resolved by the immediately following block which explains that the message only matters if the result is actually `false`:

image

michaeltlombardi:
@clairecadman, @DavidS, can we get a final approval and merge on this if there's nothing else?

DavidS (approved):

david22swan (approved):

PR #155, opened 2022-10-29
Update RSAPI spec to match actual implementation

seanmil:
The implementation of the Resource API for the `simple_get_filter` feature uses an empty array for the names paramter to signal that all resources should be returned, not `nil`. Instead of changing the code to match the spec, it probably makes more sense to change the spec to match the implementation.


seanmil:
After consideration, I think fixing the documentation/specification makes more sense than fixing the code, but if a code fix is preferred please see: https://github.com/puppetlabs/puppet-resource_api/pull/304

PR #157, opened 2023-10-26
Add Module layout specification

bastelfreak:
This document describes all files/directories that are allowed or mandatory within a module. The forge can use it as an allowlist and filter other files out. r10k/code-manager can use it as an allowlist as well when modules are pulled from version controll systems like github (similar to the `--exclude-spec` option in r10k).

### DIFF
--- a/language/module.md
+++ b/language/module.md
@@ -48,7 +48,7 @@ already explains what a module is and what it can contain.
 | `/lib/puppet/parser/functions/` | Legacy functions in Ruby |
 | `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
 | `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
-| `/functions/`         | Can contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
+| `/functions/`         | MAY contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
 | `/metadata.json`      | The `metadata.json` file MUST be present and MUST adhere to [Puppet's metadata](https://www.puppet.com/docs/puppet/latest/modules_metadata.html). [metadata-json-lint](https://github.com/voxpupuli/metadata-json-lint#metadata-json-lint) can be used to validate your file. |
 | `/README`             | A README that describes what the module does. It's best practice to add a file extension like `.md`, `.rst` when a markup language is used |
 | `/LICENSE`            | The `/LICENSE` file, with an optional file extension, SHOULD be included in the module. If the file is present, it MUST match `/metadata.json`'s license field. |
--- a/language/module.md
+++ b/language/module.md
@@ -1,0 +1,92 @@
+# Puppet Module Schema
+
+This document describes allowed (optional) and required files and directories in
+a Puppet Module Release.
+
+## Table of contents
+
+* [What is a Puppet Module](#what-is-a-puppet-module)
+* [Allowed files and directories](#allowed-files-and-directories)
+* [Explanations and reasoning](#explanations-and-reasoning)
+* [Schema versioning](#schema-versioning)
+
+## What is a Puppet Module
+
+A module in the sense of this document is a compressed tar archive. It is
+usually distributed via [forge.puppet.com](https://forge.puppet.com/). A module
+is usually developed in a git repository. Some files in the git repository are
+only used for development and testing. They should not be released.
+
+Common files often seen in a vcs repository that are used for development but
+shall not be released:
+
+`/spec`, `/Rakefile`, `/Gemfile`, `/.gitignore`, `/.github/`, `/.devcontainer`, `/Dockerfile`, `/.git`
+
+Note that above are just examples and not a complete list. The goal of this
+document is to provide an allowlist (*for a module release, not a VCS repo*),
+not a denylist.
+
+The official
+[Puppet documentation](https://www.puppet.com/docs/puppet/latest/modules_fundamentals.html)
+already explains what a module is and what it can contain.
+
+## Allowed files and directories
+
+| Directories and Files | Purpose |
+|-----------------------|---------|
+| `/manifests/`         | MAY contain Puppet code |
+| `/hiera.yaml`         | A module MAY define a Hiera configuration for Hiera data within this module |
+| `/data/`              | If the module has a `hiera.yaml`, the related data has to be within `/data/` |
+| `/templates/`         | Stores [epp](https://www.puppet.com/docs/puppet/latest/lang_template_epp.html) (preferred) or [erb](https://www.puppet.com/docs/puppet/latest/lang_template_erb.html) templates |
+| `/files/`             | Static files that Puppet code within the module will distribute |
+| `/examples/`          | Example Puppet snippets that explain how to use the module. They can be used within acceptance tests |
+| `/facts.d/`           | [External facts](https://www.puppet.com/docs/puppet/latest/external_facts.html) that are synced via [pluginsync](https://www.puppet.com/docs/puppet/latest/plugins_in_modules.html) |
+| `/lib/facter/`        | MAY contain custom facts |
+| `/lib/puppet/type/`   | Custom types |
+| `/lib/puppet/provider/` | Custom provider for one or multiple types |
+| `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
+| `/lib/puppet/parser/functions/` | Legacy functions in Ruby |
+| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
+| `/metadata.json`      | The `metadata.json` file MUST be present and MUST adhere to [Puppet's metadata](https://www.puppet.com/docs/puppet/latest/modules_metadata.html). [metadata-json-lint](https://github.com/voxpupuli/metadata-json-lint#metadata-json-lint) can be used to validate your file. |
+| `/README`             | A README that describes what the module does. It's best practice to add a file extension like `.md`, `.rst` when a markup language is used |
+| `/LICENSE`            | The `/LICENSE` file, with an optional file extension, SHOULD be included in the module. If the file is present, it MUST match `/metadata.json`'s license field. |
+| `/CHANGELOG`          | A module SHOULD contain a changelog that's updated for every release. A new release SHOULD NOT alter existing changelog entries. It MAY use a file extension if a markup language is used. The [Puppet forge](https://forge.puppet.com/) supports the markdown markup language. |
+| `/docs/`              | Directory for additional documentation |
+| `/REFERENCE.md`       | [puppet-strings](https://www.puppet.com/docs/puppet/latest/puppet_strings.html) based documentation in markdown, updated on each release |
+| `/locales/`           | Used for i18n support, can contain translated strings, deprecated within Puppet |
+| `/scripts/`           | May serve static files, like `/files/` (see [PUP-11187](https://puppet.atlassian.net/browse/PUP-11187) for background) |
+| `/tasks/`             | Contains [Tasks for Bolt](https://www.puppet.com/docs/bolt/latest/tasks.html) |
+| `/plans/`             | Contains [Plans for Bolt](https://www.puppet.com/docs/bolt/latest/plans) |
+| `/types/`             | Contains [type aliases](https://www.puppet.com/docs/puppet/latest/lang_type_aliases.html) |
+| `/bolt_plugin.json`   | The file can contain metadata about [a Bolt plugin](https://www.puppet.com/docs/bolt/latest/writing_plugins.html#module-structure) |
+
+
+## Mandatory files
+
+Mandatory are:
+* `/metadata.json`
+* `/README`
+* `/LICENSE`
+* `/CHANGELOG`
+
+## Explanations and reasoning
+
+In the past, modules sometines contained a `/Modulefile`. It contained metadata
+about the module. The `/metadata.json` is the successor. A module can now only
+have a `/metadata.json`. It must not have a `/Modulefile`.
+
+The `/REFERENCE.md` file is optional. It's generated by puppet-strings. Some
+modules might use a different tool for documentation (and then cannot generate
+a `REFERENCE.md`). If a `/REFERENCE.md` is present in the release, it has to be
+up to date.
+
+## Schema versioning
+
+This is version 0 of the schema. A changelog will be added when the schema is
+updated.
+
+A potential change is the removal of `/locales/`. The i18n support in puppet is
+currently deprecated, but still possible. When it's removed from Puppet and
+Puppetserver, the schema will be updated to reflect this.


language/module.md
```diff
@@ -0,0 +1,50 @@
...
+| `/README`             | A README that describes what the module does. It's best practice to add a suffix like `.txt`, `.md`, `.rst` for the used markup language |
+| `/LICENSE`            | `/metadata.json` specifies a license. The license text needs to be distributed in a LICENSE file. It can have a suffix if a markup language is used |
+| `/CHANGELOG`          | A Changelog that's updated for every release. A new release should not alter existing changelog entries. It can have a suffix if a markup language is used |
+| `/HISTORY`            | The history file can contain old changelogs. Useful when the tool used to generate changelogs is changed or a module is migrated to a new namespace |
+| `/docs/`              | Directory for additional documentation |
+| `/REFERENCE.md`       | [puppet-strings](https://www.puppet.com/docs/puppet/latest/puppet_strings.html) based documentation in markdown, updated on each release |
+| `/locales/`           | Referenced in https://github.com/puppetlabs/puppet/blob/945beb6ef0200b1bc7879b62131282a61fc9a935/lib/puppet/module.rb#L28, still in use? |
```
joshcooper, line 39: Yeah it can contain translated strings from a module. stdlib used to contain them, but it was removed at some point: https://github.com/puppetlabs/puppetlabs-stdlib/tree/v6.0.0/locales

Whether to enable i18n is controlled by this puppet setting https://github.com/puppetlabs/puppet/blob/5dae3b1f808d4fe0def1c5673c65273d2466ff43/lib/puppet/defaults.rb#L211

When it's `false` and the server supports the `locales` mount, then the agent will attempt to pluginsync locales from all modules:
https://github.com/puppetlabs/puppet/blob/5dae3b1f808d4fe0def1c5673c65273d2466ff43/lib/puppet/configurer/plugin_handler.rb#L33

language/module.md
```diff
@@ -0,0 +1,50 @@
...
+| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain functions written in Puppet DSL |
+| `/metadata.json`      | Metadata file that describes the module based on [a schema](https://www.puppet.com/docs/puppet/latest/modules_metadata.html) |
+| `/README`             | A README that describes what the module does. It's best practice to add a suffix like `.txt`, `.md`, `.rst` for the used markup language |
+| `/LICENSE`            | `/metadata.json` specifies a license. The license text needs to be distributed in a LICENSE file. It can have a suffix if a markup language is used |
+| `/CHANGELOG`          | A Changelog that's updated for every release. A new release should not alter existing changelog entries. It can have a suffix if a markup language is used |
+| `/HISTORY`            | The history file can contain old changelogs. Useful when the tool used to generate changelogs is changed or a module is migrated to a new namespace |
+| `/docs/`              | Directory for additional documentation |
```
bastelfreak, line 37: Some people use `/readmes/`. I think `/docs/` make more sense  but I'm fine with either.

language/module.md
bastelfreak: ah thanks, updated it.

rwaffen (approved):

language/module.md
```diff
@@ -0,0 +1,68 @@
...
+A module in the sense of this document is a compressed tar archive. It is
+usually distributed via [forge.puppet.com](https://forge.puppet.com/). A module
+is usually developed in a git repository. Some files in the git repository are
+only used for development and testing. They shall not be released.
+
+Common files often seen in a vcs repository that are used for development but
+shall not be released:
+
+* `/spec/`
+* `/Rakefile
```
tuxmea, line 17: missing backtick

language/module.md
```diff
@@ -0,0 +1,68 @@
...
+a Puppet Module.
+
+## What is a Puppet Module
+
+A module in the sense of this document is a compressed tar archive. It is
+usually distributed via [forge.puppet.com](https://forge.puppet.com/). A module
+is usually developed in a git repository. Some files in the git repository are
+only used for development and testing. They shall not be released.
+
+Common files often seen in a vcs repository that are used for development but
```
tuxmea, line 13: "This is an incomplete list as this might vary on how tests and development are done."

language/module.md
bastelfreak: done

language/module.md
bastelfreak: updated

language/module.md
tuxmea: I prefer `docs`

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+
+The official
+[Puppet documentation](https://www.puppet.com/docs/puppet/latest/modules_fundamentals.html)
+already explains what a module is and what it can contain.
+
+| Directories and Files | Purpose |
+|-----------------------|---------|
+| `/manifests/`         | contains Puppet code |
+| `hiera.yaml`          | Can define a Hiera configuration for Hiera data within this module |
+| `/data/`              | If the module has a `hiera.yaml`, the related data has to be within `/data/` |
```
ekohl, line 35: The `datadir` key in `hiera.yaml` can be set to anything. Would this also imply a modification in Puppet to enforce this? If so, should the `defaults` be preset?

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+| `/templates/`         | Stores [epp](https://www.puppet.com/docs/puppet/latest/lang_template_epp.html) or [erb](https://www.puppet.com/docs/puppet/latest/lang_template_erb.html) templates |
```
ekohl, line 36: Perhaps nudge towards EPP?
```suggestion
| `/templates/`         | Stores [epp](https://www.puppet.com/docs/puppet/latest/lang_template_epp.html) (preferred) or [erb](https://www.puppet.com/docs/puppet/latest/lang_template_erb.html) templates |
```

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+| `/lib/puppet/parser/functions/` | Legacy functions in Ruby |
+| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain functions written in Puppet DSL |
+| `/metadata.json`      | Metadata file that describes the module based on [a schema](https://www.puppet.com/docs/puppet/latest/modules_metadata.html) |
+| `/README`             | A README that describes what the module does. It's best practice to add a suffix like `.txt`, `.md`, `.rst` for the used markup language |
+| `/LICENSE`            | `/metadata.json` specifies a license. The license text needs to be distributed in a LICENSE file. It can have a suffix if a markup language is used |
+| `/CHANGELOG`          | A Changelog that's updated for every release. A new release should not alter existing changelog entries. It can have a suffix if a markup language is used |
+| `/HISTORY`            | The history file can contain old changelogs. Useful when the tool used to generate changelogs is changed or a module is migrated to a new namespace |
```
ekohl, line 52: Since you spell it out for LICENSE and CHANGELOG: this can also have a file suffix

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+| `/docs/`              | Directory for additional documentation |
+| `/REFERENCE.md`       | [puppet-strings](https://www.puppet.com/docs/puppet/latest/puppet_strings.html) based documentation in markdown, updated on each release |
```
ekohl, line 54: I still think this should be a release artifact and not be stored in git. So I'd say with [RFC2119](https://www.rfc-editor.org/rfc/rfc2119) phrasing `REFERENCE.md` MAY be stored in git.

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+| `/locales/`           | Used for i18n support, can contain translated strings |
```
ekohl, line 55: Isn't this deprecated by now? https://github.com/puppetlabs-toy-chest/puppetlabs-translate is archived in the toy chest.

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+* `/metadata.json`
+* `/README`
+* `/LICENSE`
+* `/CHANGELOG`
+* `/REFERENCE.md`
+
+In the past, modules sometines contained a `/Modulefile`. It contained metadata
+about the module. The `/metadata.json` is the successor. A module can only have
+`/Modulefile` or `/metadata.json`. For new or updated modules a `/metadata.json`
+has to be used.
```
ekohl, line 71: Current Puppet versions (Puppet 5+ IIRC, maybe even 4+) no longer understand `Modulefile`. I'd suggest to make it `metadata.json` MUST be used.

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+| `/tasks/`             | Contains [Tasks for Bolt](https://www.puppet.com/docs/bolt/latest/tasks.html) |
+| `/plans/`             | Contains [Plans for Bolt](https://www.puppet.com/docs/bolt/latest/plans) |
+| `/types/`             | Contains [type aliases](https://www.puppet.com/docs/puppet/latest/lang_type_aliases.html) |
+
+Mandatory are:
+* `/metadata.json`
+* `/README`
+* `/LICENSE`
+* `/CHANGELOG`
+* `/REFERENCE.md`
```
ekohl, line 66: I'm not sure I agree on this being mandatory.

If I look at modules I maintain then for some I just can't generate anything useful. We use rdoc style documentation headers in modules like [theforeman/foreman](https://github.com/theforeman/puppet-foreman) because our installer uses that. We rely on the parameter groups. Because of that the generated `REFERENCE.md` is unreadable.

For others where it is readable I have it in `.gitignore` and generate it at release time. That way it's guaranteed to be up to date and don't need to make sure it's in sync on every PR.

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+| `/examples/`          | Example Puppet snippets that explain how to use the module. They can be used within acceptance tests |
+| `/facts.d/`           | [External facts](https://www.puppet.com/docs/puppet/latest/external_facts.html) that are synced via [pluginsync](https://www.puppet.com/docs/puppet/latest/plugins_in_modules.html) |
+| `/lib/facter/`        | Contains custom facts |
+| `/lib/puppet/type/`   | Custom types |
+| `/lib/puppet/provider/` | Custom provider for one or multiple types |
+| `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
+| `/lib/puppet/parser/functions/` | Legacy functions in Ruby |
+| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain functions written in Puppet DSL |
```
ekohl, line 47:
```suggestion
| `/functions/`         | Can contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
```

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+| `/metadata.json`      | Metadata file that describes the module based on [a schema](https://www.puppet.com/docs/puppet/latest/modules_metadata.html) |
+| `/README`             | A README that describes what the module does. It's best practice to add a suffix like `.txt`, `.md`, `.rst` for the used markup language |
```
ekohl, line 49: In the UNIX world having a `.txt` extension is uncommon and not a markup language
```suggestion
| `/README`             | A README that describes what the module does. It's best practice to add a file extension like `.md`, `.rst` when a markup language is used |
```

language/module.md
```diff
@@ -0,0 +1,71 @@
...
+* `/spec/`
+* `/Rakefile`
+* `/Gemfile`
+* `/.gitignore`
+* `/.github/`
+* `/.devcontainer`
+* `/Dockerfile`
+
+Note that above are just examples and not a complete list. The goal of this
+document is to provide an allowlist, not a denylist.
```
ekohl, line 25: If the list of files is just an example, I'd keep it short and inline them. A listing with bullet points draws attention.

ekohl (commented):
In general I'm in favor of a specific list of what's allowed. I'd suggest the make sure [puppet-modulebuilder](https://github.com/puppetlabs/puppet-modulebuilder) has this list built in. It may be a breaking change to switch to it, so perhaps at first it should be opt-in.

Short term I think this may work for `.pmtignore`, if it follows [gitignore](https://git-scm.com/docs/gitignore):
```
# Ignore everything
*

# except
!manifest/**/*.pp
!metadata.json
!README*
!CHANGELOG*
!LICENSE*
...
```

language/module.md
bastelfreak: the current default is `/data` when `datadir` isn't set. I would like to see puppet enforcing it.

language/module.md
bastelfreak: I thought so as well last week, but I'm unsure. @joshcooper do you have an idea?

language/module.md
bastelfreak: mhm yes, will update

language/module.md
bastelfreak: good point. I updated it and the reference.md is now optional.

language/module.md
```diff
@@ -0,0 +1,70 @@
...
+
+This document describes allowed (optional) and required files and directories in
+a Puppet Module.
+
+## What is a Puppet Module
+
+A module in the sense of this document is a compressed tar archive. It is
+usually distributed via [forge.puppet.com](https://forge.puppet.com/). A module
+is usually developed in a git repository. Some files in the git repository are
+only used for development and testing. They shall not be released.
```
ekohl, line 11: If we interpret `shall` using RFC2119 then this is a prohibition. I'd not go as far.
```suggestion
only used for development and testing. They should not be released.
```

My reasoning is that on the server side you will still have old modules with those assets present, so you need to deal with them anyway.

One reason to include them is to provide everything needed to make offline changes, including tests (provided you have the other modules and dependencies too). In a distro like Fedora this is what happens for gems: they should run the test suite of that gem to verify it's working. Now granted, few Puppet modules are being packaged but it is a valid use case.

language/module.md
bastelfreak: this document is specifically not about what's in git but what's in the actual release tar ball.

> this document is specifically not about what's in git but what's in the actual release tar ball.

language/module.md
bastelfreak: thanks, updated

language/module.md
bastelfreak: but do you package the tests on fedora? on arch, we run available tests for gems during the packaging, but the tests are usually not part of the package.

language/module.md
```diff
@@ -0,0 +1,69 @@
...
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
+| `/metadata.json`      | Metadata file that describes the module based on [a schema](https://www.puppet.com/docs/puppet/latest/modules_metadata.html) |
+| `/README`             | A README that describes what the module does. It's best practice to add a file extension like `.md`, `.rst` when a markup language is used |
+| `/LICENSE`            | `/metadata.json` specifies a license. The license text needs to be distributed in a LICENSE file. It can have a suffix if a markup language is used |
+| `/CHANGELOG`          | A Changelog that's updated for every release. A new release should not alter existing changelog entries. It can have a suffix if a markup language is used |
+| `/HISTORY`            | The history file can contain old changelogs. Useful when the tool used to generate changelogs is changed or a module is migrated to a new namespace. It can have a suffix if a markup language is used |
```
ekohl, line 47: Thinking more about this: isn't `HISTORY` more of an artifact in producing `CHANGELOG`? At least for Vox Pupuli. The file in itself doesn't serve a purpose in a released tarball and shouldn't be shipped.

language/module.md
bastelfreak: good idea, updated it. we could also shorten the list.

language/module.md
bastelfreak: oh right, we append the content to CHANGELOG.md. Will update it.

language/module.md
david22swan: This was archived a long time ago.
As far as I know there are no plans to revisit it as I remember removing the compatible code for it.

bastelfreak: @joosh in https://github.com/puppetlabs/puppet-specifications/pull/157#discussion_r1373898104 mentioned the settings and at least those are still present in puppet and puppetserver. Maybe it's time to clean that up.

david22swan: Interesting, guess it was left in place....

ekohl: In my experience it's much easier to add a feature than to remove it. Finding what's missing is easy. Finding what's used only by a certain feature is much harder.

david22swan (approved):
Aside from that one comment this LGTM.
Would probably need a few more eyes on it before merging though, considering the importance.

donoghuc:
This is a good start. i would like to see `bolt_plugin.json` added as well https://www.puppet.com/docs/bolt/latest/writing_plugins.html but having this as a point of reference will be helpful.

bastelfreak:
@donoghuc thanks for the hint, I added it.

language/module.md
bastelfreak: I agree, thihs makes stfuff more complicated. We could change the spec and enforce it in `/data`.

language/module.md
```diff
@@ -0,0 +1,69 @@
...
+| `/REFERENCE.md`       | [puppet-strings](https://www.puppet.com/docs/puppet/latest/puppet_strings.html) based documentation in markdown, updated on each release |
+| `/locales/`           | Used for i18n support, can contain translated strings |
+| `/scripts/`           | Distribute a script for bolt?? |
+| `/tasks/`             | Contains [Tasks for Bolt](https://www.puppet.com/docs/bolt/latest/tasks.html) |
+| `/plans/`             | Contains [Plans for Bolt](https://www.puppet.com/docs/bolt/latest/plans) |
+| `/types/`             | Contains [type aliases](https://www.puppet.com/docs/puppet/latest/lang_type_aliases.html) |
+| `/bolt_plugin.json`   | The file can contain metadata about [a Bolt plugin](https://www.puppet.com/docs/bolt/latest/writing_plugins.html#module-structure) |
+
+Mandatory are:
+* `/metadata.json`
```
binford2k, line 57: fwiw, only `metadata.json` is actually required today.

bastelfreak: from a technical perspective, yes. But we want to enforce more files? For example the LICENSE isn't required for a module to work but without it it's really hard to use the module correctly.

language/module.md
ekohl: Perhaps something that in current Puppet can be deprecated (including warning, plus CI check) and removed in Puppet 9?

Which also implies this document should somehow be versioned. I've been a fan of the Puppet report format documentation with its explicit version numbers and "changes since previous version". Would that be a good concept to copy here?

bastelfreak: I had that in mind as well, but wasnt sure how to write it since this is the initial document.

language/module.md
ekohl: Technically you have the license field in `metadata.json`. While not sufficient for all licenses (AFAIK some license require you to provide the full license as part of deliverables), I don't see how it makes using a module correctly harder.

Again, big fan of RFC2119 definitions for these kinds of specifications. For license I'd say:

The `LICENSE` file, with an optional file suffix, SHOULD be included in the module. If the file is present, it MUST match `metadata.json`'s `license` field.

And in the same trend for `metadata.json`:

The `metadata.json` file MUST be present and MUST adhere to [Puppet's metadata](https://www.puppet.com/docs/puppet/8/modules_metadata.html). [metadata-json-lint](https://rubygems.org/gems/metadata-json-lint) can be used to validate your file.

language/module.md
ekohl: Puppet's report format was initially version 0. I'd make this version 0, formalizing what's already out there in the world. Then Puppet 9 could formalize version 1.

bastelfreak: This is a good idea to move forward. I added a section about format and versioning.

language/module.md
bastelfreak: We should keep it for now because it's still possible to use it. I added a note that the module schema will be updated to a new version and `/locales/` removed when it's actually dropped from puppet and puppetserver.

language/module.md
bastelfreak: great idea, updated it with your suggestions.

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+The official
+[Puppet documentation](https://www.puppet.com/docs/puppet/latest/modules_fundamentals.html)
+already explains what a module is and what it can contain.
+
+## Allowed files and directories
+
+| Directories and Files | Purpose |
+|-----------------------|---------|
+| `/manifests/`         | contains Puppet code |
+| `hiera.yaml`          | Can define a Hiera configuration for Hiera data within this module |
```
ekohl, line 38: If you want to be consistent
```suggestion
| `/hiera.yaml`          | Can define a Hiera configuration for Hiera data within this module |
```

ekohl (commented):
I wonder if it's useful to show a tree form.

```console
$ mkdir -p docs examples facts.d files functions lib/facter lib/puppet/{type,provider,functions,parser/functions,puppet_x,augeas/lenses} locales plans scripts tasks templates types
$ touch CHANGELOG LICENSE README REFERENCE.md bolt_plugin.json hiera.yaml metadata.json
$ tree
.
├── bolt_plugin.json
├── CHANGELOG
├── docs
├── examples
├── facts.d
├── files
├── functions
├── hiera.yaml
├── lib
│   ├── facter
│   └── puppet
│       ├── augeas
│       │   └── lenses
│       ├── functions
│       ├── parser
│       │   └── functions
│       ├── provider
│       ├── puppet_x
│       └── type
├── LICENSE
├── locales
├── manifests
├── metadata.json
├── plans
├── README
├── REFERENCE.md
├── scripts
├── tasks
├── templates
└── types

24 directories, 7 files
```

On the console it helpfully shows directories as blue and files as white so it's even easier to understand.

I'd even consider a separate chapter for each file/directory describing it. Some descriptions are already quite long, but many of them could be further enhanced with references to official documentation. For example, `files` can explain the `puppet://` mechanism, but also the `file()` function. And that the Puppetserver will serve the files as a fileserver. Similarly, for lib which things are synced to agents.

Ramesh7:
Just a thought : as we have consider bolt scenario here so should we consider below 2 files also to be part of specification ?
- Puppetfile
- bolt-project.yaml

donoghuc:
@Ramesh7 no those are for bolt projects. A bolt project is not a module. This spec is for modules.

jordanbreen28 (approved):

Ramesh7 (approved):

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
+| `/metadata.json`      | The metadata.json file MUST be present and MUST adhere to [Puppet's metadata](https://www.puppet.com/docs/puppet/latest/modules_metadata.html). [metadata-json-lint](https://github.com/voxpupuli/metadata-json-lint#metadata-json-lint) can be used to validate your file. |
+| `/README`             | A README that describes what the module does. It's best practice to add a file extension like `.md`, `.rst` when a markup language is used |
+| `/LICENSE`            | The `/LICENSE` file, with an optional file suffix, SHOULD be included in the module. If the file is present, it MUST match `/metadata.json`'s license field. |
+| `/CHANGELOG`          | A Changelog that's updated for every release. A new release should not alter existing changelog entries. It can have a suffix if a markup language is used |
```
ekohl, line 55:
```suggestion
| `/CHANGELOG`          | A module SHOULD contain a changelog that's updated for every release. A new release SHOULD NOT alter existing changelog entries. It MAY use a file extension if a markup language is used. The Puppet forge supports the markdown markup language. |
```

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+The official
+[Puppet documentation](https://www.puppet.com/docs/puppet/latest/modules_fundamentals.html)
+already explains what a module is and what it can contain.
+
+## Allowed files and directories
+
+| Directories and Files | Purpose |
+|-----------------------|---------|
+| `/manifests/`         | contains Puppet code |
+| `/hiera.yaml`          | Can define a Hiera configuration for Hiera data within this module |
```
ekohl, line 38:
```suggestion
| `/hiera.yaml`         | A module MAY define a Hiera configuration for Hiera data within this module |
```

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+
+The official
+[Puppet documentation](https://www.puppet.com/docs/puppet/latest/modules_fundamentals.html)
+already explains what a module is and what it can contain.
+
+## Allowed files and directories
+
+| Directories and Files | Purpose |
+|-----------------------|---------|
+| `/manifests/`         | contains Puppet code |
```
ekohl, line 37: The `manifests` directory module MAY contain Puppet code

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+| `/facts.d/`           | [External facts](https://www.puppet.com/docs/puppet/latest/external_facts.html) that are synced via [pluginsync](https://www.puppet.com/docs/puppet/latest/plugins_in_modules.html) |
+| `/lib/facter/`        | Contains custom facts |
+| `/lib/puppet/type/`   | Custom types |
+| `/lib/puppet/provider/` | Custom provider for one or multiple types |
+| `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
+| `/lib/puppet/parser/functions/` | Legacy functions in Ruby |
+| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
+| `/metadata.json`      | The metadata.json file MUST be present and MUST adhere to [Puppet's metadata](https://www.puppet.com/docs/puppet/latest/modules_metadata.html). [metadata-json-lint](https://github.com/voxpupuli/metadata-json-lint#metadata-json-lint) can be used to validate your file. |
```
ekohl, line 52:
```suggestion
| `/metadata.json`      | The `metadata.json` file MUST be present and MUST adhere to [Puppet's metadata](https://www.puppet.com/docs/puppet/latest/modules_metadata.html). [metadata-json-lint](https://github.com/voxpupuli/metadata-json-lint#metadata-json-lint) can be used to validate your file. |
```

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+| Directories and Files | Purpose |
+|-----------------------|---------|
+| `/manifests/`         | contains Puppet code |
+| `/hiera.yaml`          | Can define a Hiera configuration for Hiera data within this module |
+| `/data/`              | If the module has a `hiera.yaml`, the related data has to be within `/data/` |
+| `/templates/`         | Stores [epp](https://www.puppet.com/docs/puppet/latest/lang_template_epp.html) (preferred) or [erb](https://www.puppet.com/docs/puppet/latest/lang_template_erb.html) templates |
+| `/files/`             | Static files that Puppet code within the module will distribute |
+| `/examples/`          | Example Puppet snippets that explain how to use the module. They can be used within acceptance tests |
+| `/facts.d/`           | [External facts](https://www.puppet.com/docs/puppet/latest/external_facts.html) that are synced via [pluginsync](https://www.puppet.com/docs/puppet/latest/plugins_in_modules.html) |
+| `/lib/facter/`        | Contains custom facts |
```
ekohl, line 44:
```suggestion
| `/lib/facter/`        | MAY contain custom facts |
```

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+| `/lib/puppet/provider/` | Custom provider for one or multiple types |
+| `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
+| `/lib/puppet/parser/functions/` | Legacy functions in Ruby |
+| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
+| `/metadata.json`      | The metadata.json file MUST be present and MUST adhere to [Puppet's metadata](https://www.puppet.com/docs/puppet/latest/modules_metadata.html). [metadata-json-lint](https://github.com/voxpupuli/metadata-json-lint#metadata-json-lint) can be used to validate your file. |
+| `/README`             | A README that describes what the module does. It's best practice to add a file extension like `.md`, `.rst` when a markup language is used |
+| `/LICENSE`            | The `/LICENSE` file, with an optional file suffix, SHOULD be included in the module. If the file is present, it MUST match `/metadata.json`'s license field. |
```
ekohl, line 54:
```suggestion
| `/LICENSE`            | The `/LICENSE` file, with an optional file extension, SHOULD be included in the module. If the file is present, it MUST match `/metadata.json`'s license field. |
```

ekohl (commented):
Thoughts on adding more RFC2119 style descriptions to further explain what to expect within a module?

joshcooper:
> Thoughts on adding more RFC2119 style descriptions to further explain what to expect within a module?

:+1: from me. Technically puppet accepts an empty module directory:

```
$ mkdir /home/josh/.puppetlabs/etc/code/environments/production/modules/foo

$ cat mod.rb 
require 'puppet'
Puppet.initialize_settings
Puppet::Util::Log.newdestination(:console)
Puppet::ApplicationSupport.push_application_context(Puppet::Util::RunMode[:user])
envs = Puppet.lookup(:environments)
env = envs.get!('production')
puts Puppet::Module.find('foo', env)

$ bx ruby mod.rb                                                          
Module foo(/home/josh/.puppetlabs/etc/code/environments/production/modules/foo)
```

I believe the only strict requirement is the module name must satisfy https://github.com/puppetlabs/puppet/blob/49d74c94c848d4d4f4d022b58ff8135b2f46e7b0/lib/puppet/module.rb#L481

If the module contains `metadata.json` (which it should), then it must contain these keys https://github.com/puppetlabs/puppet/blob/49d74c94c848d4d4f4d022b58ff8135b2f46e7b0/lib/puppet/module.rb#L253 and the `dependencies` key must contain a (possible empty) list of modules it depends on.

language/module.md
```diff
@@ -0,0 +1,92 @@
+# Puppet Module Schema
+
+This document describes allowed (optional) and required files and directories in
+a Puppet Module Release.
```
natemccurdy, line 4: What is the definition of a "Puppet Module Release"?

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+| `/README`             | A README that describes what the module does. It's best practice to add a file extension like `.md`, `.rst` when a markup language is used |
+| `/LICENSE`            | The `/LICENSE` file, with an optional file extension, SHOULD be included in the module. If the file is present, it MUST match `/metadata.json`'s license field. |
+| `/CHANGELOG`          | A module SHOULD contain a changelog that's updated for every release. A new release SHOULD NOT alter existing changelog entries. It MAY use a file extension if a markup language is used. The [Puppet forge](https://forge.puppet.com/) supports the markdown markup language. |
+| `/docs/`              | Directory for additional documentation |
+| `/REFERENCE.md`       | [puppet-strings](https://www.puppet.com/docs/puppet/latest/puppet_strings.html) based documentation in markdown, updated on each release |
+| `/locales/`           | Used for i18n support, can contain translated strings, deprecated within Puppet |
+| `/scripts/`           | Distribute a script for bolt?? |
```
natemccurdy, line 59: If this is still a question, don't merge this.

What is the purpose of `/scripts/`?

language/module.md
bastelfreak: the .tar.gz archive of a module, that you can download from the forge for example.

language/module.md
bastelfreak: I think I saw `/scripts/` here: https://github.com/WhatsARanjit/puppet-node_manager/tree/master/scripts
. This isn't really specific to a puppet module but a script that's related to the module content? I think having this in a module is okay, the question is in which directory.

language/module.md
natemccurdy: That's good context for readers of this document, and I think the definition of "Puppet Module Release" should be noted here in the document.

I see there's some confusion about if this is the spec for what's in git vs. not, etc... So defining the scope of this document up front could help to reduce confusion and clarify what exactly is being defined.

bastelfreak: That's at the top, the `## What is a Puppet Module`. Do you think that's not sufficient?

natemccurdy: 🤦 Ah, I somehow missed that section entirely, sorry. The "What is a Puppet Module" section looks good to me.

language/module.md
joshcooper: The directory is correct, see https://puppet.atlassian.net/browse/PUP-11187 and https://github.com/puppetlabs/puppet/blob/fce49baa3767fffb000f1b26d3a7b4edbec58999/lib/puppet/file_serving/configuration.rb#L91

bastelfreak: thanks, I updated the comment with a link to the jira issue.

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+
+The official
+[Puppet documentation](https://www.puppet.com/docs/puppet/latest/modules_fundamentals.html)
+already explains what a module is and what it can contain.
+
+## Allowed files and directories
+
+| Directories and Files | Purpose |
+|-----------------------|---------|
+| `/manifests/`         | MAY contains Puppet code |
```
gcoxmoz, line 37: nitpick: 'may contain', no 's'

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+| `/examples/`          | Example Puppet snippets that explain how to use the module. They can be used within acceptance tests |
+| `/facts.d/`           | [External facts](https://www.puppet.com/docs/puppet/latest/external_facts.html) that are synced via [pluginsync](https://www.puppet.com/docs/puppet/latest/plugins_in_modules.html) |
+| `/lib/facter/`        | MAY contain custom facts |
+| `/lib/puppet/type/`   | Custom types |
+| `/lib/puppet/provider/` | Custom provider for one or multiple types |
+| `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
+| `/lib/puppet/parser/functions/` | Legacy functions in Ruby |
+| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
+| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
+| `/functions/`         | Can contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
```
jstraw, line 51: if one if these is MAY, we probably need to be more clear for the rest as to the fact that they do not have to exist if you're not doing them.

ekohl:
```suggestion
| `/lib/facter/`        | MAY contain custom facts |
| `/lib/puppet/type/`   | Custom types |
| `/lib/puppet/provider/` | Custom provider for one or multiple types |
| `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
| `/lib/puppet/parser/functions/` | Legacy functions in Ruby |
| `/lib/puppet_x/`      | Custom Ruby modules to extend types, providers, functions or facts |
| `/lib/augeas/lenses/` | Custom [Augeas](https://augeas.net/) lenses |
| `/functions/`         | MAY contain [functions written in Puppet DSL](https://www.puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html) |
```

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+have a `/metadata.json`. It must not have a `/Modulefile`.
+
+The `/REFERENCE.md` file is optional. It's generated by puppet-strings. Some
+modules might use a different tool for documentation (and then cannot generate
+a `REFERENCE.md`). If a `/REFERENCE.md` is present in the release, it has to be
+up to date.
+
+## Schema versioning
+
+This is version 0 of the schema. A changelog will be added when the schema is
```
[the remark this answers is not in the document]
bastelfreak: at other places Puppet only used major versions for schema. For example for the report format. I'm fine with everything here.

language/module.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Allowed files and directories
+
+| Directories and Files | Purpose |
+|-----------------------|---------|
+| `/manifests/`         | MAY contain Puppet code |
+| `/hiera.yaml`         | A module MAY define a Hiera configuration for Hiera data within this module |
+| `/data/`              | If the module has a `hiera.yaml`, the related data has to be within `/data/` |
+| `/templates/`         | Stores [epp](https://www.puppet.com/docs/puppet/latest/lang_template_epp.html) (preferred) or [erb](https://www.puppet.com/docs/puppet/latest/lang_template_erb.html) templates |
+| `/files/`             | Static files that Puppet code within the module will distribute |
```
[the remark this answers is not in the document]
joshcooper: nit I think you mean `s/filebucket/file serving/'

Filebucket is a different thing/concept with its own set of issues, but the `files` directory is what puppetserver serves files from and it does have issues when large numbers of files.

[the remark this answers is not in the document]
bastelfreak: ah yes, will update it.

joshcooper (approved):

language/module.md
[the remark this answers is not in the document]
bastelfreak: mhm I think a recommendation here is tricky. "you should not add too many files"? I'm afraid that we add something like this we will start a discussion about what "too many" is.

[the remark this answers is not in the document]
joshcooper: The main problem is fileserving is implemented in Ruby and so each `file_{metadata,content}` request requires a JRuby instance which competes with catalog, etc requests. So lots of little files are bad. Recursive requests can make the bad situation worse. Also note pluginsync is implemented as a recursive file request. And large files are bad too, as puppet's fileserving wasn't designed for that scale. There was a proposal to move fileserving to clojure, eg https://puppet.atlassian.net/browse/SERVER-2035, but was not implemented. Our recommendation is if you have many files, then create a package and use puppet to manage the package.

PR #158, opened 2024-04-24
Add missing paths to allowed module layout

seanmil:


language/module.md
```diff
@@ -45,6 +45,12 @@ already explains what a module is and what it can contain.
 | `/lib/puppet/type/`   | Custom types |
 | `/lib/puppet/provider/` | Custom provider for one or multiple types |
 | `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
+| `/lib/puppet/datatypes/` | Custom Puppet Datatypes |
```
bastelfreak, line 48: is that the same as `/lib/puppet/type/`?

seanmil: No, `/lib/puppet/type/` is for Resource Types. `/lib/puppet/datatypes/` is for custom data types.  "Type" is overused in Puppet, but ah well.  You can see an example here: https://github.com/hlindberg/tahu/tree/master/lib/puppet/datatypes/tahu

bastelfreak: ahhh those types! :D

seanmil: I pushed an update which clarifies "Resource types" in the type and provider lines.

language/module.md
```diff
@@ -42,9 +42,15 @@ already explains what a module is and what it can contain.
 | `/examples/`          | Example Puppet snippets that explain how to use the module. They can be used within acceptance tests |
 | `/facts.d/`           | [External facts](https://www.puppet.com/docs/puppet/latest/external_facts.html) that are synced via [pluginsync](https://www.puppet.com/docs/puppet/latest/plugins_in_modules.html) |
 | `/lib/facter/`        | MAY contain custom facts |
-| `/lib/puppet/type/`   | Custom types |
-| `/lib/puppet/provider/` | Custom provider for one or multiple types |
+| `/lib/puppet/type/`   | Custom Resource types |
+| `/lib/puppet/provider/` | Custom provider for one or multiple Resource types |
 | `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
+| `/lib/puppet/datatypes/` | Custom Puppet Data types |
```
joshcooper, line 48: I think you're thinking of `/types` like https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/types/absolutepath.pp?

language/module.md
```diff
@@ -42,9 +42,15 @@ already explains what a module is and what it can contain.
 | `/examples/`          | Example Puppet snippets that explain how to use the module. They can be used within acceptance tests |
 | `/facts.d/`           | [External facts](https://www.puppet.com/docs/puppet/latest/external_facts.html) that are synced via [pluginsync](https://www.puppet.com/docs/puppet/latest/plugins_in_modules.html) |
 | `/lib/facter/`        | MAY contain custom facts |
-| `/lib/puppet/type/`   | Custom types |
-| `/lib/puppet/provider/` | Custom provider for one or multiple types |
+| `/lib/puppet/type/`   | Custom Resource types |
+| `/lib/puppet/provider/` | Custom provider for one or multiple Resource types |
```
joshcooper, line 46: Technically, modules can put whatever they want in the `/lib` directory and it will be packaged in the module and pluginsynced to agents. So from an allowlist perspective, listing individual `/lib/puppet/*` directories is misleading. But from a "what are the various extension points" perspective, it's informative.

bastelfreak: > Technically, modules can put whatever they want in the /lib directory
is that something that we should allow here? Do you have anything in mind that isn't already covered by the different `/lib/puppet/` subdirs?

language/module.md
joshcooper: ah this is answered in https://github.com/puppetlabs/puppet-specifications/pull/158#discussion_r1577858919

language/module.md
joshcooper: > is that something that we should allow here?

I think we should allow anything under `/lib` since that's how puppet behaves today. For example, there was some early confusion about `/lib/puppetx` vs `/lib/puppet_x`. And given `/lib/augeas` there may be some others?

In the future, I think it would be good to document which extension points are allowed in the `/lib` directory and add a deprecation warning to if we detect a dir/file that isn't allowed during pluginsync. Then in a major release, only pluginsync files that are in the allow list.

language/module.md
```diff
@@ -42,9 +42,15 @@ already explains what a module is and what it can contain.
...
 | `/lib/puppet/functions/` | Modern functions in Ruby for the new API |
+| `/lib/puppet/datatypes/` | Custom Puppet Data types |
+| `/lib/puppet/face/`   | Custom Puppet Faces |
+| `/lib/puppet/feature/` | Custom Puppet Features for providers |
+| `/lib/puppet/property/` | Custom Puppet Properties for Types/Providers |
```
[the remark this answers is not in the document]
joshcooper: Also `/lib/puppet/applications` and `/lib/puppet/reports`

language/module.md
bastelfreak: > add a deprecation warning to if we detect a dir/file that isn't allowed during pluginsync

Do you think it's realistic to introduce the warnings in Puppet 8 or maybe with Puppet 9.0?

Issue #159, opened 2024-06-18
Add a "tags" field to task metadata


hsnodgrass:
## Use Case

When developing Bolt tasks, it would be nice to be able to add arbitrary string tags to tasks for purposes of organization, automation, etc. For example, when writing Bolt tasks that map to compliance rules, it would be nice to be able to add tags with the rule names to the task metadata. These tags wouldn't hold any semantic meaning to Bolt or the task itself, but they would be an easy way for people to identify which tasks map to which compliance rules.

## Describe the Solution You Would Like

I would like to see Bolt task metadata support an optional `tags` field that accepts an array of strings, with no default value. That way, arbitrary tags could be assigned to a task in the metadata like this:

```json
{
  "description": "Some task",
  "input_method": "stdin",
  "tags": [
    "arbitrary_tag1",
    "Arbitrary Tag 2"
  ]
}
```

## Describe Alternatives You've Considered

I'm currently using code comments for this type of information, but those are difficult to parse for automation and differ in style between implementing languages.

davidsandilands:
transferring to puppet-specifications as mentioned above

bastelfreak:
I like this idea and I think it makes sense.

hsnodgrass:
Opened https://github.com/puppetlabs/puppet-specifications/pull/161 to address this.

donoghuc:
How is this meant to be consumed? Would a task runner be expected to surface this information in every place `description` is currently surfaced? Is a task runner responsible for providing this context when running a task?

hsnodgrass:
My thought was that is up to the individual task runner if they want to implement special handling for these tags. My personal use case is to query the tags from the metadata files with my own tooling to help generate documentation. I just want a tags key to be recognized as valid in a Task metadata file when it comes to linting, validation, etc.

donoghuc:
What is preventing you from doing that today? Is there a schema check in a task runner or part of the module ecosystem that is complaining about the addition?

hsnodgrass:
In my team's workflow, as far as I'm aware, there aren't any checks against the metadata schema. However, because we end up shipping the tasks to our customers as part of the SCE modules, I'd prefer to only use officially documented parameters in the metadata. I don't know what kind of validation workflows customers might have, and I don't want the potential of a support ticket coming in because we added an arbitrary key to task metadata. Admittedly, the chances of that happening are low, but it also seems like the cost of adding a `tags` key to the metadata schema is low as well.

[...]

PR #161, opened 2024-06-20
Add identifiers field to Bolt task metadata schema

hsnodgrass:
This commit adds the `identifiers` field to Bolt task metadata. A brief description of why an `identifiers` field would be useful is included in the `tasks/README.md` update. A more concrete example is writing Bolt tasks to help with security compliance. In order to map tasks to compliance rules, you either need to name your task after a rule or use comments in the implementation script. With the new `identifiers` field, you could add a keys under `identifiers` showing details about the security controls implemented.

```json
"identifiers": {
  "compliance_controls": {
    "cis": ["some_control", "another_one"],
    "stig": ["V-000000"]
  }
}
```

This PR addresses issue #159

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -84,6 +84,7 @@ The preferred style of the keys should be `snake_case`.
 
 **remote**: Default `false`. All implementation of this task operate on a remote target using the `_target` metaparam.(rev 4)
 
+**identifiers**: An arbitrary map of key-value pairs without inherent, semantic meaning to a task runner. Intended to provide identification and / or categorization of tasks that can be consumed by custom or third-party tooling.
 
 #### Example Task Metadata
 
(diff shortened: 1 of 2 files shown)


donoghuc:
Rolling out something like this needs to happen in the task spec and have a path forward for all task runners (most notably PE).

hsnodgrass:
> Rolling out something like this needs to happen in the task spec and have a path forward for all task runners (most notably PE).

Is this something I can help with? If so, where should I look?

tasks/README.md
```diff
@@ -84,6 +84,7 @@ The preferred style of the keys should be `snake_case`.
 
 **remote**: Default `false`. All implementation of this task operate on a remote target using the `_target` metaparam.(rev 4)
 
+**tags**: An array of strings used for arbitrary tags without inherent, semantic meaning to Bolt. Useful for categorizing tasks.
```
donoghuc, line 87: We shouldn't mention Bolt here. This spec is for any task runner.

hsnodgrass: Fixed

hsnodgrass:
@donoghuc I've updated this PR based on the discussion in #159

donoghuc (approved):

hsnodgrass:
@donoghuc anything else need to be done on this before it gets merged?

donoghuc:
nope, looks good to me. I think we've waited long enough to allow for other opinions.

